PR #5, opened 2017-08-27
Use chip UID as USB serial number

florolf:
Instead of hardcoding the serial number in the firmware, use the
device's manufacturer-provided UID as the serial number.

This makes it easier to distinguish between multiple adaptors plugged
into the same host without having to build a separate firmware for each
one.


normaldotcom:
@florolf do you know of a method to see which interface (can0, can1, etc) corresponds to which serial number? I haven't found an easy way of doing this. Thanks!

daniel-santos:
I guess this is the iSerial field in lsusb or called "serial" under /sys.  I haven't played with this one yet, but mine shows up as 000000000001 -- I presume the default until you change it?

PR #12, opened 2018-07-29
Feature/nucleo f072rb board

vondraussen:
This adds the [Nucleo F072RB](https://www.st.com/content/st_com/en/products/evaluation-tools/product-evaluation-tools/mcu-eval-tools/stm32-mcu-eval-tools/stm32-mcu-nucleo/nucleo-f072rb.html) as a target in Makefile and config.h
Also adds the possibility to give each board a different USBD_MANUFACTURER_STRING.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,13 +33,14 @@ DEP     += $(ASM_OBJ:%.asmo=%.d)
 ELF = build/$(BOARD)/gsusb_$(BOARD).elf
 BIN = bin/gsusb_$(BOARD).bin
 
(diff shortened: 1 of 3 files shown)


HubertD:
Thanks for your contribution, @vondraussen!

Unfortunately, I'm not convinced that I want explicit support for some generic dev board that doesn't even have a CAN transceiver. Am I missing something?

vondraussen:
You are right, I'll keep it as a fork and update the readme how to setup the board.
However, thanks for the great project!

PR #18, opened 2019-06-03
Fix enable_irq interpretation of primask

ndushman:
The logic for enable_irq is backwards; it should enable interrupts when primask is 0, not 1.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -34,7 +34,7 @@ inline int disable_irq(void) {
 }
 
 inline void enable_irq(int primask) {
-    if (primask)
+    if (!primask)
         asm volatile("cpsie i\n");
 }
 


HubertD:
Hi,
sorry for being unresponsive for so long. I'll try to do better in the future, let's see if this works out.

Your patch sounds reasonable, though it makes me scratch my head and ask "How could this ever possibly have worked, then?"

Anyway, I'd rather try to clean up this mess, so that it's not so much of a headache.

ndushman:
It works because every call to a queue operation toggles whether interrupts are enabled. There's a call to a queue operation in the main loop https://github.com/candle-usb/candleLight_fw/blob/53b4b915af166575d7c93d63811001ae89e45e33/src/main.c#L101 So when an interrupt occurs, it could take an extra cycle of the main loop before the interrupt is processed, but overall it does work.

HubertD:
Thanks for the clarification.
Merging this as it's an obvious fix;
Reworked the enable/disable_irq() functionality in my current development branch because the status quo was obviously hard to understand.

daniel-santos:
> The logic for enable_irq is backwards; it should enable interrupts when primask is 0, not 1.

Yikes!  I never looked closely at this.  Thank you so much for this fix!!!!

PR #23, opened 2019-10-29
nexulm fixes

fenugrec:
I have rebased some (all ?) of the commits that had been added to @nexulm's fork.

It affects the Makefile, so when the cmake branch is ready/merged some of this will need to be carried over (just some extra board types).

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -69,9 +69,12 @@ int main(void)
 
 	gpio_init();
 
+#if BOARD == BOARD_canable
+	led_init(&hLED, LED1_GPIO_Port, LED1_Pin, true, LED2_GPIO_Port, LED2_Pin, true);
+#else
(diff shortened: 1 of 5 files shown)


src/main.c
```diff
@@ -1,244 +1,247 @@
...
+	if(!frame)
+	  return;
+	
+	if (USBD_GS_CAN_SendFrame(&hUSB, frame) == USBD_OK) {
+	        queue_push_back(q_frame_pool, frame);
+	} else {
+	        queue_push_front(q_to_host, frame);
+	}
+}
+
```
jhofstee: this changes line endings of the whole file. Which is fine by itself, but should be a different patch, since this makes the whole diff unreadable. I have cleanup patches for that.. Perhaps that should go in first?

fenugrec:
merge in 68c514e8fcb0e682e0e06752bd9cf6c4c09d4ea6

PR #24, opened 2019-10-29
LED functions : support GPIO pins >= 8

fenugrec:
This required an uint16 arg to pass the pin #.
I made this commit for debugging on an stm32F072-discovery board, as well as a custom board to be added soon, where some LEDs are wired to GPIO8 and 9.

This is fairly low-impact. AFAIK the binary size changes by less than 6 bytes (smaller or larger depending on board type).

### DIFF
--- a/include/led.h
+++ b/include/led.h
@@ -49,7 +49,7 @@ typedef struct {
 
 typedef struct {
 	void* port;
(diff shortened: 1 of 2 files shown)


fenugrec:
merged in 0434e8e

PR #25, opened 2019-10-30
re-organize everything to a cmake project structure

HubertD:
This PR moves almost all files (except the ones in src/) to match a halfway modern cmake structure.

I consider something like this a must as I'm fed up with Makefiles and eclipse as an IDE.
Having cmake enables the use of IDEs like vscode, qtcreator and clion.

I can compile, link and run the cmake project on a STM32F072 discovery board;
USB-Part seems to work well, however I don't have the setup at hand to test CAN functionality.
Would be great if someone could verify everything still works at least as well as it did before.

There is another PR in the queue (branch update_st_libs) that builds on this PR and updates all ST libraries (HAL and USB Device lib) and as well throws out a lot of unneeded third party code.
Don't have a replacement (yet) for ST USB Device lib and it's silly license, though :-(


fenugrec:
Some comments :

#### -Werror
On arm-none-eabi-gcc 9.2 , having the -Werror option causes the build to fail with a few lines of
` error: listing the stack pointer register 'sp' in a clobber list is deprecated`
coming from the STM32_HAL CMSIS code. Suggested fix : 
1. remove -Werror until the HAL code is updated, or
2. (better) add -Wno-deprecated to the compile flags, until HAL code is updated

#### flash target
There is no longer a "flash-${target}" shortcut to run dfu-util, but adding this to CmakeLists.txt seems to do the trick : (I added the -R flag to have dfu-util reset the target)
```
function(add_flash_target target)
	add_custom_target(
		flash-${target} dfu-util -d 1d50:606f -a 0 -R -s 0x08000000 -D ${target}.bin
	)
endfunction()

....
add_executable(candleLight_fw ${SOURCE_FILES})
 ...
add_flash_target(candleLight_fw)

```

I'm currently testing and it appears to brick my stm32f042-based device - it doesn't enumerate and I need to reset into the builtin USB-DFU bootloader.

libs/STM32_HAL/CMakeLists.txt
```diff
@@ -0,0 +1,113 @@
...
+    include/cmsis
+    include/stm32f0xx
+    include/cmsis/device
+    config/
+)
+
+add_library(STM32_HAL_STM32F042x6 OBJECT ${SOURCES} src/cmsis/startup_stm32f042x6.S)
+    target_include_directories(STM32_HAL_STM32F042x6 PUBLIC ${INCLUDE_DIRS})
+    target_compile_options(STM32_HAL_STM32F042x6 PRIVATE -Wno-unused-parameter)
+    target_compile_definitions(STM32_HAL_STM32F042x6 PUBLIC STM32F072xB)
```
fenugrec: Aha ! this needs to be
`target_compile_definitions(STM32_HAL_STM32F042x6 PUBLIC STM32F042x6)`

my thing still doesn't enumerate though. WIP

HubertD: Hrm. Same with my candleLight with STM32F072. Does not enumerate.
Usb works, though, with my update_st_libs branch...

HubertD:
canfdtest works with the update_st_libs branch.
maybe I should drop this one and directly go to update_st_libs?

fenugrec:
Yeah, I think that would be ok - if canfdtest works on your f072 target, it should work on f042 maybe requiring only the tweaks you added today. I'll test your update_st_libs in a few minues

normaldotcom:
Haven't gotten a chance to compile/test this on 042 yet, but the cmake setup looks super clean.

 Is there any particular way you have tested functionality/throughput in the past? It would be interesting to have some sort of scripted test that can be run to verify that things are working properly--I've made some Python scripts to do some testing, but mostly for testing packet tx/rx and not actual throughput.

fenugrec:
At my end, F042 fails to compile on the update_st_libs branch, due to too much RAM usage (!? something changed in the linker scripts wrt heap size I think, it's on my TODO list to check. Which makes me want to statically allocate the FIFOs and get rid of the heap, but that will be a different PR !)

canfdtest with two devices gives a sanity check but no performance info. We might be able to script something with some of the can-utils tools though.

HubertD:
> due to too much RAM usage
Entirely possible, but interestingly doesn't happen on my machine (tm).
You're building with a GCC9 toolchain, right? Which one? Custom-built?
[my linker scripts base on the ones from the gnu-arm-gcc toolchain, come from my libreucpp project and leave, besides stack and heap, 2.5K RAM on F042, and 4K on F072]

I'd link to see malloc() getting banned from the firmware, but you're right, that's for later.
And I really have to dig into linker scripts sooner or later, as I think HEAP_SIZE should not be statically assigned, but calculated from RAM_SIZE-STACK_SIZE-(all statically allocated stuff)

HubertD:
As for a quick fix, reducing __STACK_SIZE and/or __HEAP_SIZE in the F042 linker script should do the trick, though I don't understand where the rest of the RAM goes.

HubertD:
> canfdtest with two devices gives a sanity check but no performance info. We might be able to script something with some of the can-utils tools though.

Don't have the equipment at home right now, but running canfdtest together with canbusload gives (at high bitrates) a good indication on performance imho.

fenugrec:
> You're building with a GCC9 toolchain, right? Which one? Custom-built?
Just 9.2.0, pre-packaged from my distro
https://www.archlinux.org/packages/community/x86_64/arm-none-eabi-gcc/

Can you paste the size summary of your build of candlelight ? After reducing __HEAP_SIZE, @ 11debf4 (update_st_libs), I get
```
   text	   data	    bss	    dec	    hex	filename
  12648	    528	   2184	  15360	   3c00	candleLight_fw
```
the bss section is huge; there's a few monster structs that take up the most of that 2k:
```
$ arm-none-eabi-nm candleLight_fw --print-size --size-sort | grep " b " -i
...
address    size       
200009ec 0000008c b _global_atexit0
200007c8 00000200 b USBD_StrDesc
20000290 00000274 b hpcd_USB_FS
20000504 000002c4 b hUSB
```
interestingly, @ f24a409 (only the cmake PR) , the usage looks like
```
20000150 00000200 B USBD_StrDesc
20000350 00000220 B hpcd_USB_FS
200005bc 00000224 B hUSB
```
Those USB structs took on a lot of weight in the update !

So, with update_st_libs + minor tweaks, it runs on F042, enumerates properly. But canfdtest only works for a short while before failing with data mismatches (??). 

Back to this PR : I found out why the vectors don't end up in the binary: libs/STM32_HAL/src/cmsis/startup_stm32f042x6.S is not being compiled at all. It has to do with ASM not being "enabled" for the project, I'm looking into it.

fenugrec:
Ah ! Can you try this, on this branch without update_st_libs : 
change root CMakeLists.txt  project() to
`project(candleLightFirmware C ASM)`

seems to work here. If that works for you, then we can merge this, then continue working on udpate_stlibs

HubertD:
adding ASM to project languages seems to work on my F072 discovery, though I can't test CAN right now.
@fenugrec if it work's for you, let's merge this one!

PR #27, opened 2019-11-02
Update st libs, remove unneeded third-party code

HubertD:
Update STM32 HAL and USB Device Library to latest CubeMX versions.
Use newlib-nano nosys library instead of custom syscall stubs.
Reduce RAM waste to compensate for larger STM USB lib structs.


HubertD:
@fenugrec regarding your off-topic comment on the cmake PR https://github.com/candle-usb/candleLight_fw/pull/25#issuecomment-548862395 :

I moved all USB descriptor data from DATA to ROM section.
Since the usb device library and/or the USB peripheral seem to depend on descriptor data in RAM, they are copied to a common buffer when requested.

sizes now look like this:
```
text	   data	    bss	    dec	    hex	filename
12704	    192	   1724	  14620	   391c	candleLight_fw
```
not perfect, but much better.

HubertD:
Still, I think RAM usage must be quite high on STM32F42.

Having CAN_QUEUE_SIZE=64, there are some serious allocations going on:
64 x 24 bytes = 1536 bytes "frame pool"
3 queues à (16 + 64 x 4) bytes = 816 bytes
Another estimated 128 bytes go for a heap instance of USBD_GS_CAN_HandleTypeDef.
With malloc data structures, I expect some 2.5KB of HEAP usage.
That together with 2K static ram usage (DATA + BSS) leaves about 1.5K headroom for stack.

Should be enough, but it's pretty close. 
Maybe we should lower CAN_QUEUE_SIZE at least on the F042 devices.

Numbers would btw. explain data corruptions you saw when static RAM allocation was significantly higher than it's now.

HubertD:
@fenugrec do you still see the "_global_atexit0" structure in your BSS section?

With my gcc8 toolchain, I don't have it in the binary and I see no reason why it should be there.

We could probaby overwrite the weak-linked newlib function "void __register_exitproc(void) { }":
https://stackoverflow.com/questions/48711221/how-to-prevent-inclusion-of-c-library-destructors-and-atexit

But I think it should just not be there in the first place :)

daniel-santos:
CMake isn't my favorite (I prefer autotools), but I'm so glad to hear about the conversion -- it really needed it!  Hopefully I'll get to play with this soon, but I'll probably wait until you tag a release. :)

fenugrec:
I've been merging a lot of stuff to master and creating tons of conflict - I'll take care of them when we're ready to merge this, since it's mostly my fault.

Yes, `_global_atexit0` is still in my builds for some reason: pulled in by a call to "atexit" at the end of _start().

> data corruption

I'm not sure how low heap space could cause this kind of weird behavior though:
```
[.....]
Databyte 7 mismatch !
expected: 0078: [8] 1e 1f 20 21 22 23 24 25
received: 0078: [8] 1f 20 21 22 23 24 25 26

Test messages sent and received: 542
```
Nice data, but shifted by 1 byte ? but not all the time ? 
Hmmm... is it possible some frames are dropped, so that the correct frame never makes it to the receiver ?

fenugrec:
Ah ! it's even more interesting than that, and possibly not a problem after all.
Here's the traffic dump while running `canfdtest -g can1 -l 20` (running 20 loops, I had to run it a dozen times before getting a failure).

```
  can0  077   [8]  0A 0B 0C 0D 0E 0F 10 11
  can0  078   [8]  0B 0C 0D 0E 0F 10 11 12
  can0  077   [8]  0B 0C 0D 0E 0F 10 11 12
  can0  077   [8]  0C 0D 0E 0F 10 11 12 13
  can0  077   [8]  0D 0E 0F 10 11 12 13 14
  can0  078   [8]  0C 0D 0E 0F 10 11 12 13
  can0  078   [8]  0D 0E 0F 10 11 12 13 14
  can0  078   [8]  0E 0F 10 11 12 13 14 15
  can0  077   [8]  0E 0F 10 11 12 13 14 15
  can0  078   [8]  0F 10 11 12 13 14 15 16
  can0  077   [8]  0F 10 11 12 13 14 15 16
```

So the host sends "077  ...", and the DUT responds with ID 078 and incrementing all data bytes by 1. At some point the host has time to send 3 packets in a row (those starting with data 0B, 0C, 0D) before the DUT responds. Seems like that confuses canfdtest, because all replies are in there !

HubertD:
@fenugrec I'm curious, what is your test setup?
Last time I checked with canfdtest (PC <-> PeakCAN USB <- can -> candleLight <-> PC), I had no such issues.

fenugrec:
My setup is PC->cannette-> can->cannette->PC (two identical devices, just another F042 variant)

HubertD:
okay, finally have a test setup here.
Using a candleLight vs. a PCAN-USB FD, I see no problems with canfdtest running for several minutes, both with candleLight as the generator or the responder.

HubertD:
@fenugrec 

>   can0  077   [8]  0A 0B 0C 0D 0E 0F 10 11
>   can0  078   [8]  0B 0C 0D 0E 0F 10 11 12
>   can0  077   [8]  0B 0C 0D 0E 0F 10 11 12
>   can0  077   [8]  0C 0D 0E 0F 10 11 12 13
>   can0  077   [8]  0D 0E 0F 10 11 12 13 14
>   can0  078   [8]  0C 0D 0E 0F 10 11 12 13
>   can0  078   [8]  0D 0E 0F 10 11 12 13 14

What seems odd to me is that the response to "077   [8]  0B..." is seen even before the transmit is acknowledged. That should not be the case, and I suppose that's what also gets canfdtest confused?

We probably have a problem in the candleLight firmware here, but I would suppose it's not related to the changes in this branch..? Sure it doesn't also happen in current master?

HubertD:
also tested with two candleLights, canfd runs forever without errors.
works for me (tm).

fenugrec:
> Sure it doesn't also happen in current master?

Sorry, I've just now checked at three different points (current master 9e493849 , this PR d09bf69, and a previous master (11debf4c), and all have the same canfdtest problem within a few seconds. At least it's easy to test, failures occur quickly. So the problem is not caused by the lib update, it seems.

About your changes in here:
```       USBD_Init(&hUSB, (USBD_DescriptorsTypeDef*)&FS_Desc, DEVICE_FS);```

I notice here you explicitly drop 'const' from FS_Desc; it's probably safe (at least for now) since the ST libs don't modify that struct. I wish ST had marked that argument as 'const', but that's ST.
Maybe we should add a warning to that line in case the ST libs change behavior ? I think the uC would hardfault anyway but I haven't tested.

fenugrec:
Oops ! I half-accidentally merged this to master. It mostly works, although with a reduced throughput...

fenugrec:
Closing this , since after further testing everything seems to be operating as it was pre-merge.

Issue #32, opened 2019-11-09
/usr/bin/objcopy: Unable to recognise the format of the input file `candleLight_fw'


brandonros:
```
 CC='/usr/bin/arm-none-eabi-gcc' CFLAGS="--specs=nosys.specs" CMAKE_OBJCOPY="/usr/lib/arm-none-eabi/bin/objcopy" cmake .
make VERBOSE=1
```

```
/usr/bin/cmake -S/home/brandon/candleLight_fw -B/home/brandon/candleLight_fw --check-build-system CMakeFiles/Makefile.cmake 0
/usr/bin/cmake -E cmake_progress_start /home/brandon/candleLight_fw/CMakeFiles /home/brandon/candleLight_fw/CMakeFiles/progress.marks
make -f CMakeFiles/Makefile2 all
make[1]: Entering directory '/home/brandon/candleLight_fw'
make -f libs/STM32_HAL/CMakeFiles/STM32_HAL_STM32F072xB.dir/build.make libs/STM32_HAL/CMakeFiles/STM32_HAL_STM32F072xB.dir/depend
make[2]: Entering directory '/home/brandon/candleLight_fw'
cd /home/brandon/candleLight_fw && /usr/bin/cmake -E cmake_depends "Unix Makefiles" /home/brandon/candleLight_fw /home/brandon/candleLight_fw/libs/STM32_HAL /home/brandon/candleLight_fw /home/brandon/candleLight_fw/libs/STM32_HAL /home/brandon/candleLight_fw/libs/STM32_HAL/CMakeFiles/STM32_HAL_STM32F072xB.dir/DependInfo.cmake --color=
make[2]: Leaving directory '/home/brandon/candleLight_fw'
make -f libs/STM32_HAL/CMakeFiles/STM32_HAL_STM32F072xB.dir/build.make libs/STM32_HAL/CMakeFiles/STM32_HAL_STM32F072xB.dir/build
make[2]: Entering directory '/home/brandon/candleLight_fw'
make[2]: Nothing to be done for 'libs/STM32_HAL/CMakeFiles/STM32_HAL_STM32F072xB.dir/build'.
make[2]: Leaving directory '/home/brandon/candleLight_fw'
[ 23%] Built target STM32_HAL_STM32F072xB
[...]

PR #35, opened 2019-11-11
fix CMakeLists + explicit LED polarity settings

fenugrec:
I failed to test the build on the last PR merge and broke cmake due to a syntax error (missing ").

Also inverts the LED logic for the canalyze board, and sets explicitly the polarity for all board types.

@brandonros , please test your device against this

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,7 +96,7 @@ endfunction()
 
 ########## generate list of targets.
 # the "_fw" part is appended automatically
-set(TGT042_LIST "cantact" "canalyze" canable" "usb2can")
+set(TGT042_LIST "cantact" "canalyze" "canable" "usb2can")
 set(TGT072_LIST "candleLight")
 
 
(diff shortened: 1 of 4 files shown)


brandonros:
@fenugrec I will test, but can you advise here? https://github.com/candle-usb/candleLight_fw/issues/32#issuecomment-552128547

Looking for a fix so that I do not need to manually override `objcopy` binary path.

brandonros:
@fenugrec we kind of flipped what is happening. now, when device is plugged in: LEDs are off. then, I call resetDevice / setupDevice (I wrote WebUSB interface for this, by the way) and both green + red LED stay solid.

brandonros:
Eh, this is probably good enough. Green stays steady when the device is up, and red blinks on transmission. I say send it through.

fenugrec:
webUSB... 
Yes, I agree the LED behavior is a bit strange but I don't think it's specific to your device. I have another patch to address this. In any case the logic should be correct now  for  all boards.

PR #36, opened 2019-11-11
Update README.md

brandonros:


fenugrec:
please modify commands to build from a separate directory - compiling in-tree is never recommended. mkdir build, cd build, cmake .. -D...
Also maybe a comment that editing the cmake file may be required if the binary names / paths don't match

fenugrec:
merged, thanks

PR #39, opened 2019-11-30
Adds Travis-Ci and deploy on tags

vondraussen:
This would add continuous integration and deploys if a version is tagged.
Every target will be available as its own tarball in the releases section.

The repo owner would need to change the section:
```
api_key:
    secure: <encrypted token>
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,37 @@
+dist: bionic
+language: generic
+
+install:
+  - wget https://github.com/Kitware/CMake/releases/download/v3.14.5/cmake-3.14.5-Linux-x86_64.sh
+    && chmod +x cmake-3.14.5-Linux-x86_64.sh
+  - mkdir -p /opt/cmake-3.14.5 && ./cmake-3.14.5-Linux-x86_64.sh --skip-license --prefix=/opt/cmake-3.14.5
+  - PATH=/opt/cmake-3.14.5/bin:$PATH
+  - wget https://armkeil.blob.core.windows.net/developer/Files/downloads/gnu-rm/8-2019q3/RC1.1/gcc-arm-none-eabi-8-2019-q3-update-linux.tar.bz2
+  - tar xf gcc-arm-none-eabi-8-2019-q3-update-linux.tar.bz2 -C /opt
+  - mkdir build && cd build
+  - cmake .. -DCMAKE_TOOLCHAIN_FILE=../cmake/gcc-arm-none-eabi-8-2019-q3-update.cmake
+  - make
+
+before_deploy:
+  - tar czvf canable_fw_${TRAVIS_TAG}.tar.gz canable_fw*
+  - tar czvf canalyze_fw_${TRAVIS_TAG}.tar.gz canalyze_fw*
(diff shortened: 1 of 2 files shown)


fenugrec:
Nice ! Thanks for working on this.

What is the reason for installing cmake-3.15 manually, isn't there already a version bundled in the basic 'bionic' image ?

> api_key
what goes in there, a github repo api key ? I've never configured a CI setup before.

vondraussen:
The cmake version (3.12.?) which comes with bionic is a bit dated and lacks the "add_link_options" feature.

Regarding the api_key:

1. you need to link the repo with https://travis-ci.org
2. you need to create an repo access token for travis
   - via [github](https://youtu.be/BFpSD2eoXUk?t=414) <-- that is probably a bit easier if you don't want to install travis-cli
   - via [travis-cli](https://docs.travis-ci.com/user/deployment/releases/#using-travis-ci-client-to-populate-initial-deployment-configuration) <-- I did it this way

vondraussen:
@fenugrec shall I change the .travis.yml so that a ENV variable is used? I guess that would be better because you would only need to add the variable in your travis ui for this repo. No change on the .travis.yml needed on your side.

fenugrec:
> The cmake version (3.12.?)  ... lacks the "add_link_options"

Ah, thanks for pointing that out, I just updated our cmake_minimum_required() line.

>  change the .travis.yml so that a ENV variable is used?

Sorry, I haven't had time to read up on how these pieces fit together. At first glance it sounds like a good idea instead of hardcoding the key in the repo, but I'm not sure of the implications.

Also I was looking at the release packaging, I think ours would preferably include all compiled images instead of one-per-file ? We only have had one official release so far which was essentially  the whole repo with source and precompiled images.

I think it might also make more sense as more targets are added, if we don't need to manually add them to the yml file as well. I'm thinking maybe we can leverage cmake's CPack here, I'll have to look into it.

vondraussen:
> Also I was looking at the release packaging, I think ours would preferably include all compiled images instead of one-per-file ? We only have had one official release so far which was essentially the whole repo with source and precompiled images.

Good point!

> I think it might also make more sense as more targets are added, if we don't need to manually add them to the yml file as well. I'm thinking maybe we can leverage cmake's CPack here, I'll have to look into it.

I have not much experience with cmake's cpack but I can take a look into it and update the pull request.

fenugrec:
I used cpack a few years ago to setup another project,
https://github.com/fenugrec/freediag/blob/master/CMakeLists.txt#L210

It works OK, so I should be able to do something similar here. Then I assume we'd need to add "make package" and a bit of glue somewhere in the .yml

vondraussen:
@fenugrec the cpack feature seems to be more useful if you have a real application which is installed to an OS. I cant find an easy way to create a deploy package with cpack.
What I did instead is, packing all *_fw* files into one archive, as you requested. As long as the naming convention will not be changed, it should work without change (.travis.yml) for new targets.
Source code is already part of the deployments.

Example:
img

fenugrec:
Looks great !
Yeah, globbing *_fw* is probably OK for a while.

I don't think I have permissions to integrate travis in this project, but I can merge the PR at least..

PR #41, opened 2020-01-09
Flip endpoints

brandonros:
Why do we receive to OUT and transmit to IN?...


fenugrec:
Because IN and OUT refer to USB endpoints where "OUT" means "host-to-device". 
The stm32 USB peripheral docs also use the same terminology.
And the "0x81" address needs that 0x80 bit to identify it as an IN endpoint in its descriptor. See the USB2 spec, 9.6.6 .

PR #45, opened 2020-04-08
Clarify stm32f0x2 support

brandonros:
image

https://www.st.com/content/ccc/resource/sales_and_marketing/promotional_material/brochure/0f/e0/12/6f/fe/20/44/5a/brstm32f0.pdf/files/brstm32f0.pdf/jcr:content/translations/en.brstm32f0.pdf

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # candleLight_gsusb
 [![Build Status](https://travis-ci.org/candle-usb/candleLight_fw.svg?branch=master)](https://travis-ci.org/candle-usb/candleLight_fw)
 
-This is a firmware for stm32f0-based USB-CAN adapters, notably:
+This is a firmware for stm32f0x2-based USB-CAN adapters, notably:
 - candleLight: https://github.com/HubertD/candleLight
 - cantact: http://linklayer.github.io/cantact/
 - canable (cantact clone): http://canable.io/
--- a/README.md
+++ b/README.md
@@ -1,12 +1,12 @@
 # candleLight_gsusb
 [![Build Status](https://travis-ci.org/candle-usb/candleLight_fw.svg?branch=master)](https://travis-ci.org/candle-usb/candleLight_fw)
 
-This is a firmware for stm32f0x2-based USB-CAN adapters, notably:
-- candleLight: https://github.com/HubertD/candleLight
-- cantact: http://linklayer.github.io/cantact/
-- canable (cantact clone): http://canable.io/
-- USB2CAN: https://github.com/roboterclubaachen/usb2can
-- CANAlyze: https://kkuchera.github.io/canalyze/
+This is a firmware for STM32F042x6/STM32F072xB-based USB-CAN adapters, notably:
+- candleLight: https://github.com/HubertD/candleLight (STM32F072xB)
+- cantact: http://linklayer.github.io/cantact/ (STM32F042x6)
+- canable (cantact clone): http://canable.io/ (STM32F042x6)
+- USB2CAN: https://github.com/roboterclubaachen/usb2can (STM32F042x6)
+- CANAlyze: https://kkuchera.github.io/canalyze/ (STM32F042x6)
 
 It implements the interface of the mainline linux gs_usb kernel module and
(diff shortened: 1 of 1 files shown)


fenugrec:
Looks good, but I think in theory the firmware could fit on F0x2x4 (such as F042C4), that last '4' indicates 16kB flash and IIRC we're around 15000-15200.

You mentioned there's no premade devices currently available; if you're interested I have some extra boards of a variant I made (F042C6T6; optionally opto-isolated). I was hoping to make more but the current context is of course far from ideal ...

brandonros:
> Looks good, but I think in theory the firmware could fit on F0x2x4 (such as F042C4), that last '4' indicates 16kB flash and IIRC we're around [15000-15200](tel:15000-15200).
> 
> You mentioned there's no premade devices currently available; if you're interested I have some extra boards of a variant I made (F042C6T6; optionally opto-isolated). I was hoping to make more but the current context is of course far from ideal ...

What were you going to sell them for? A perfect device would be in this form factor: 

ELM327 FORScan OBD2 Adapter ELMconfig OBDII USB Scanner Diagnostic Scan Tool with MS-CAN HS-CAN Switch for Ford F150 F250 Cars and Light Trucks Code Reader Diagnose on Windows Only https://www.amazon.com/dp/B07MQ8GHG3/ref=cm_sw_r_cp_api_i_urDJEb2DFN6X6

OBD-II -> USB, easy for entry level consumers/hobbyists to get involved without needing to deal with pigtail cables/wires

they’d sell like hotcakes at $20-$25/shipping because they can be advertised as “able to do car variant coding from your web browser, no linux” because I ported portions of the gs_usb driver to userspace

fenugrec:
> What were you going to sell them for?

Not sure... I thought maybe selling as kits or through some china-based fab/PCBA place. <25$ shipped would be completely impossible in both cases.

> A perfect device would be in this form factor:

For my own uses, OBD is a completely separate topic. This one has DB9 / terminal block connections only:
IMG_20200408_095224684c

 
> ELM327

For various reasons, I don't want anything to do with ELM327 clones and have no mind to make yet another crappy clone. And there's especially no reason to even try if they're available for so cheap...

brandonros:
we are on the same page about elm327, which is why i am trying to say “let’s
get a good + easy to use stm32fx0 option on the market”

it’s all about form factor

car enthusiasts that aren’t programmers go cross-eyed when you ask them to buy pigtail cables

fenugrec:
I would think the most cost-effective option for that would be to buy a "mini-VCI J2534" clone and reflash the mcu if it's not locked. But IIRC they use an old LPC part so this firmware wouldn't work as-is.

fenugrec:
Back on topic; squashed + merged in https://github.com/candle-usb/candleLight_fw/commit/28a31e6fd3f7a575ed35047d76c8f6c56aeafb20 . Thanks !

brandonros:
last thing: https://www.aliexpress.com/item/672381461.html?spm=a2g0o.productlist.0.0.37614d16Xk0pFC&algo_pvid=14aed250-8efc-491c-a455-be87e55f5757&algo_expid=14aed250-8efc-491c-a455-be87e55f5757-1&btsid=0ab6f81615863922687115238e294e&ws_ab_test=searchweb0_0,searchweb201602_,searchweb201603_

can you clarify why j2534 devices are typically mentioned with a manufacturer? like a device only claims to support j2534 toyota... i thought OBD + CAN + j2534 was the same independent of manufacturers?

fenugrec:
> can you clarify why j2534 devices are typically mentioned with a manufacturer?

I'm not entirely sure. Part of the reason for cheap clones like that is simply to be able to reuse the OEM dll (AFAIK those miniVCI clones are kindof a scam for that now because they have very old  buggy firmware only compatible with old drivers, something like that.)

And maybe partly because of certain optional J2534 features only present on some devices? Or just how much the OEM software is locked to a certain type of device / driver ? They're allowed to do that of course but in some countries (including USA) the minimum requirement is to allow reflashing with any compliant J2534 tools, as I understand. They could be total dicks and prevent generic tools to be used for all other diag / servicing fonctionality.

PR #51, opened 2020-07-02
Reset CAN on USB suspend. Fixes #46.

### LINKED ISSUE #46: canable device hangs after the host computer reboot
Hi All,

I have a very strange issue. I am using a canbale clone hardware with candlelight firmware. The clone is a part of another pcb and the STM32F0 is powered separately in stead of usb 5v. Therefore while using the device communicating with another can device, if the host computer reboots, the canbale device gets hanged. The device is correctly getting configured on the host computer. But I cannot send/receive any data until I reset the power of STM32F0. 

The probable solution is to power the STM32F0 by USB 5v which is now not possible as the STM32F0 is part of another board. My question is is there a way to reset the STM32F0 from the USB side from the HOST computer? or Is this the expected behavior?

normaldotcom:
This fixes issue #46, where the host system reboots or restarts the USB port while the microcontroller running this firmware continues running.

This is accomplished by disabling all CAN ports when the USB Suspend event occurs. This particular implementation is not my favorite (modifying usbd_conf.c) but I couldn't come up with a cleaner way to implement this.

### DIFF
--- a/src/usbd_gs_can.c
+++ b/src/usbd_gs_can.c
@@ -349,6 +349,24 @@ void USBD_GS_CAN_SetChannel(USBD_HandleTypeDef *pdev, uint8_t channel, can_data_
 	}
 }
 
+// Handle USB suspend event
+void USBD_GS_CAN_SuspendCallback(USBD_HandleTypeDef  *pdev)
+{
+	// Disable CAN and go off bus on USB suspend
+	USBD_GS_CAN_HandleTypeDef *hcan = (USBD_GS_CAN_HandleTypeDef*) pdev->pClassData;
+
+	if(hcan != NULL)
+	{
+		for(uint32_t i=0; i<NUM_CAN_CHANNEL; i++)
+			if(hcan->channels[i] != NULL)
+				can_disable(hcan->channels[i]);
+	}
+
+	if(hcan != NULL && hcan->leds != NULL)
+		led_set_mode(hcan->leds, led_mode_off);
+}
+
+
 static led_seq_step_t led_identify_seq[] = {
 	{ .state = 0x01, .time_in_10ms = 10 },
 	{ .state = 0x02, .time_in_10ms = 10 },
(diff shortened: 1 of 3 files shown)


src/usbd_gs_can.c
```diff
@@ -349,6 +349,24 @@ void USBD_GS_CAN_SetChannel(USBD_HandleTypeDef *pdev, uint8_t channel, can_data_
...
+	USBD_GS_CAN_HandleTypeDef *hcan = (USBD_GS_CAN_HandleTypeDef*) pdev->pClassData;
+
+	if(hcan != NULL)
+	{
+		for(uint32_t i=0; i<NUM_CAN_CHANNEL; i++)
+			if(hcan->channels[i] != NULL)
+				can_disable(hcan->channels[i]);
+	}
+
+	if(hcan != NULL && hcan->leds != NULL)
```
fenugrec, line 365: hehe.. why not add this inside the `if (hcan...)` block right above ?
Or even, invert the logic and just start the funcction with `if (hcan == NULL) return;` and have one less indentation !

src/usbd_gs_can.c
```diff
@@ -349,6 +349,24 @@ void USBD_GS_CAN_SetChannel(USBD_HandleTypeDef *pdev, uint8_t channel, can_data_
...
+// Handle USB suspend event
+void USBD_GS_CAN_SuspendCallback(USBD_HandleTypeDef  *pdev)
+{
+	// Disable CAN and go off bus on USB suspend
+	USBD_GS_CAN_HandleTypeDef *hcan = (USBD_GS_CAN_HandleTypeDef*) pdev->pClassData;
+
+	if(hcan != NULL)
+	{
+		for(uint32_t i=0; i<NUM_CAN_CHANNEL; i++)
+			if(hcan->channels[i] != NULL)
```
fenugrec, line 361: while I'm nitpicking on style, I think I'd like explicit  { } braces even though there's just one line.

xdaco:
@fenugrec  This is awesome news. I will test it asap.

Thanks a ton !!

xdaco:
@fenugrec  I have tested it and the fix is working. Great job !!

fenugrec:
good to know. You have to thank normaldotcom not me !
I'll try to merge this "soon"

fenugrec:
Thanks for looking at this and sorry I never merged it. I did merge PR #94 , essentially a superset of this PR that addresses some of the other causes of the intermittent problems.

PR #52, opened 2020-07-02
Update CANable-specific LED initialization

normaldotcom:
This commit adds blinking on startup for CANable devices (could enable for all boards if desired). It also inverts the behavior of the power LED to better indicate when the device is powered on as CANable devices do not have a dedicated power LED.

This also fixes a typo in the CANable DFU USB string.


fenugrec:
merged in 465e957db9adf238272315decdf4a83751e7f138 thanks ! I'll test here on a different device (not publicly added yet) and probably enable for all devices.

PR #54, opened 2020-09-20
Add Copperforge VulCAN

WardBenjamin:
This adds the Copperforge VulCAN (https://shop.copperforge.cc/products/ac41) to the supported device list in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@ This is firmware for certain STM32F042x/STM32F072xB-based USB-CAN adapters, nota
 - canable (cantact clone): http://canable.io/ (STM32F042x6)
 - USB2CAN: https://github.com/roboterclubaachen/usb2can (STM32F042x6)
 - CANAlyze: https://kkuchera.github.io/canalyze/ (STM32F042x6)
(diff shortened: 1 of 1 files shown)


fenugrec:
merged, thanks !

fritz-net:
```
It implements the CAN physical layer as defined in ISO 11898-2:2016 and SAE J2284-1 to SAE J2284-5. This implementation enables reliable communication in the CAN FD fast phase at data rates up to 5 Mbit/s, meaning that it is compatible with all cars which use CAN high-speed.

The VulCAN is firmware-compatible with the CANtact and CANable projects, and uses one of two open-source firmware options: CANtact slcan-compatible (cantact-fw) or candleLight (candleLight_fw).

Note: The VulCAN is a CAN high-speed device. If you are interested in an equivalent CAN low-speed device, please let us know.
```

I'm just wondering if i miss the capabilities of CandleLight / stm32f0*2 devices for Can FD or if the product page is wrong.

WardBenjamin:
Good catch, I'll get in touch and see what I can do. I think you're right.

Issue #29, opened 2019-11-05
[RFC] add git commit hash to USB serial / product string


fenugrec:
I'm working with 2 devices and jumping accross many branches, rebasing, etc, and I'm easily confused as to which build my devices are actually running.

I'm thinking it might be nice to add a short git hash (as produced by `git rev-parse --short HEAD` to one of the USB strings, but I'm not sure which one

- iSerial is already the built-in ID of the chip, and I think it's not the right place for a fw version anyway
- append to iProduct ? ugly, and irrelevant..
- add another USB string ID ? haven't looked into this, might be the way to go...
- some other mechanism ?
- is it even a good idea ?

HubertD:
unfortunately don't have a good idea where to put a software version, either.
it'd be nice to have, though.
Using iSerial is not a good option imho, as it would probably be used for udev rules to assign a certain can interface name to a certain can bus. That shouldn't change with a software update.
Using iProduct might not be great, but I see it as an option. At least, it would be easily accessible...

normaldotcom:
I append git rev/repo information to the product string on the stock CANable firmware. It's certainly not pretty, but it works well. Example dmesg output:

[1263955.892535] usb 1-1: New USB device strings: Mfr=1, Product=2, SerialNumber=3
[1263955.892538] usb 1-1: Product: CANable 1205aa6 https://github.com/normaldotcom/cantact-fw
[1263955.892541] usb 1-1: Manufacturer: Protofusion Labs
[1263955.892543] usb 1-1: SerialNumber: 0024000D524E431120373533

brian-brt:
I'll add a :+1: for putting a git hash somewhere. A [vendor-specific descriptor](https://stackoverflow.com/questions/19464269/where-to-find-information-on-custom-usb-descriptor-types) which just has the string seems like a "clean" approach, but pretty annoying to retrieve.

It's been a while since I dealt with USB devices on Windows, but I think there's a possibility of it being unhappy if you change any of the standard strings without also changing bcdDevice. Ignoring the new value and caching the old one seems more unlikely. Have you done any experiments with that @normaldotcom?

fenugrec:
Just working on this right now ! Trying iConfiguration in the config descriptor (seems like the code was ready for this, config.h already had USBD_CONFIGURATION_STRING_FS so I just needed to add USBD_IDX_CONFIG_STR in the config descriptor).
It's pretty straightforward : lsusb -v will show the string that way. 

[...]

PR #60, opened 2020-12-01
Always queue frames to the host in order

brian-brt:
The other logic could sometimes send frames out of order. I only
observed it with test code that sends lots of frames really fast, but it
could happen with normal traffic too.

I don't see why the logic was different in the first place either. The
main loop executes pretty fast, so always putting the frame in the queue
makes all the logic simpler.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -220,20 +220,8 @@ void SystemClock_Config(void)
 
 bool send_to_host_or_enqueue(struct gs_host_frame *frame)
 {
-	if (USBD_GS_CAN_GetProtocolVersion(&hUSB) == 2) {
-		queue_push_back(q_to_host, frame);
-		return true;
-
-	} else {
-		bool retval = false;
-		if ( USBD_GS_CAN_SendFrame(&hUSB, frame) == USBD_OK ) {
-			queue_push_back(q_frame_pool, frame);
-			retval = true;
-		} else {
-			queue_push_back(q_to_host, frame);
-		}
-		return retval;
-	}
+	queue_push_back(q_to_host, frame);
+	return true;
 }
 
 void send_to_host()


brian-brt:
@fenugrec I wonder if this might fix the out-of-order problem you found at https://github.com/candle-usb/candleLight_fw/pull/27#issuecomment-552088421. Do you still have your setup handy to try it?

fenugrec:
Hi, ooh, that would be nice. I do have the same setup, even better now (3 identical devices !). I'm going to hook them all up and re-test.

`USBD_GS_CAN_GetProtocolVersion(&hUSB) == 2`
I see this has to do with whether or not timestamps are enabled; can you elaborate on why it makes a difference for the proposed changes  ?

brian-brt:
I don't see why this code should work differently with or without timestamping in the first place. I think that's just a historical artifact.

f4f47116 added this new version of the relevant code, but only used it when timestamping is enabled. My wild guess is enabling timestamps revealed the out-of-orderness, so this bug got fixed, but for some reason @HubertD decided to preserve the old behavior when timestamping isn't enabled.

fenugrec:
Hehe sorry, I was looking "backwards" and thought you were adding that if / else construct. I'll check again when my brain works properly : )

fenugrec:
Ok. Hooked up two devices (same computer) and ran some tests, `canfdtest -g can1` and `canfdtest -v can0` on the other.
- let it run a good while, and manually stopped it after >1E6  messages (no error)
- swapped roles of can0 and can1 : occasionally fails (after 1500 -  4500 messages, with the same out-of-sequence problem)

Good, so I can still repro the issue. But I've been swapping around 4 devices with different firmware builds  (really need to look at #29 again ) and I've totally lost track. Perhaps slightly different builds explains why one direction seems to never fail.

Running out of time just now but later today I'll try your patch.

Note on "protocol_version" 2 : that's set if  GS_CAN_MODE_HW_TIMESTAMP is given, but current kernels ( https://elixir.bootlin.com/linux/v5.9.11/source/drivers/net/can/usb/gs_usb.c ) don't even have that defined ? perhaps it was for a future version that wasn't merged...

brian-brt:
Oops, not what I was trying to do...

fenugrec:
>  I wonder if this might fix the out-of-order problem you found at #27

Yes, seems like it did the trick. cangen has been running for a few minutes (around 48% bus load according to canbusload), where it took no more than a few seconds to fail before.


I can't think of a reason not to merge this, unless it was to preserve compatibility with the original hardware gs_usb was written for ? then again that hardware was probably running this firmware anyway, so...

fenugrec:
Merged in 8b3a7b4 , reworded commit message to refer to the issue comments. Thanks !

Issue #59, opened 2020-11-24
Support for berr-reporting


brian-brt:
I've got a project with a CANable clone, and I'd like to get information about what kind of CAN errors it sees. SocketCAN supports this via the berr-reporting (aka `CAN_CTRLMODE_BERR_REPORTING`) feature. I'm planning to hook that up.

The `CAN_ESR` register on this hardware has a `LEC` field which exposes this information. I've worked with the c_can Linux driver before, which uses hardware with a very similar interface for bit errors, so I plan to follow that pattern. I'm planning to expose that information from this firmware, and add support for gs_usb to expose it via the existing SocketCAN APIs. It looks like the existing supported features / enabled features bitmasks should make this easy to implement in a fully backwards-compatible way on both sides.

I don't see a mailing list or anything, so I'm opening this issue to discuss. Is there another place for that? Does anybody here know of any similar or conflicting work? Any suggestions/comments on the order of sending out the changes here and on the kernel side?

fenugrec:
I think this is, at the moment, the best place for this kind of discussion. It's a fairly quiet project; the creator @HubertD is busy on other things and I'm only trying to assist in my own limited time.

I currently don't have much t o say on this, but there is one "major" change coming "soon" (tm) - I'm supposed to merge the hang on suspend PR #51 (I really wanted to be able to reproduce the issue though), then the ST SPL update PR #27 which is already a complicated merge. So, fair warning - if you start implementing this and I merge 27 before you finish, you may have a slightly messy rebase to take care of : )

brian-brt:
I went to look at the place to add this, and found that it's already there. Turns out candleLight currently reports bit errors unconditionally already... gs_usb doesn't allow turning the flag on though, which was the thing that confused me. I don't have a need to fix that. Just need to make sure the rest of my software isn't too clever about looking for that flag. Changing it in a backwards-compatible way would be a bit tricky too (old candleLight = always does it even when the flag is off, but new candleLight = won't report bit errors without flipping the flag?).

I do see some improvements to report more of the information exposed by the CAN peripheral though. Should have a PR for that ready soon.

fenugrec:
ah nice. Error reporting is an area I'm not very familiar with. I'm all for improving the fw; I wonder how many applications would be "broken" by candle not being backwards-compatible ? And would the required "fix" be anything more than trivially enabling the proper mode when bringing the i/f up  ?

brian-brt:
[...]

PR #61, opened 2020-12-02
Better bit error reporting

brian-brt:
It reports more information with more accurate ordering with respect to
other frames sent and received.

Fixes candle-usb/candleLight_fw#59

### DIFF
--- a/src/can.c
+++ b/src/can.c
@@ -232,8 +232,18 @@ uint32_t can_get_error_status(can_data_t *hcan)
 	return can->ESR;
 }
 
-bool can_parse_error_status(uint32_t err, struct gs_host_frame *frame)
+static bool status_is_active(uint32_t err)
 {
+	return !(err & (CAN_ESR_BOFF | CAN_ESR_EPVF));
+}
+
+bool can_parse_error_status(uint32_t err, uint32_t last_err, can_data_t *hcan, struct gs_host_frame *frame)
+{
+	/* We build up the detailed error information at the same time as we decide
+	 * whether there's anything worth sending. This variable tracks that final
+	 * result. */
+	bool should_send = false;
+
 	frame->echo_id = 0xFFFFFFFF;
 	frame->can_id  = CAN_ERR_FLAG | CAN_ERR_CRTL;
 	frame->can_dlc = CAN_ERR_DLC;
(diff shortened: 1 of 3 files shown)


src/can.c
```diff
@@ -246,51 +256,99 @@ bool can_parse_error_status(uint32_t err, struct gs_host_frame *frame)
 	frame->data[6] = 0;
 	frame->data[7] = 0;
 
-	if ((err & CAN_ESR_BOFF) != 0) {
+	if (err & CAN_ESR_BOFF) {
 		frame->can_id |= CAN_ERR_BUSOFF;
+		if (!(last_err & CAN_ESR_BOFF)) {
```
fenugrec, line 261: part of me wants to write `should_send = (err ^ last_err) & CAN_ERR_BOFF;` but it's probably clearer the way you have it P)

src/can.c
```diff
@@ -232,8 +232,18 @@ uint32_t can_get_error_status(can_data_t *hcan)
 	return can->ESR;
 }
 
-bool can_parse_error_status(uint32_t err, struct gs_host_frame *frame)
+static bool status_is_active(uint32_t err)
 {
+	return !(err & (CAN_ESR_BOFF | CAN_ESR_EPVF));
+}
+
+bool can_parse_error_status(uint32_t err, uint32_t last_err, can_data_t *hcan, struct gs_host_frame *frame)
```
fenugrec, line 240: perhaps we should have a short comment somewhere (here or in the .h) to clarify that err and last_err are simply a copy of the CAN_ESR register, with error counts in the two upper bytes ?

src/can.c
```diff
@@ -246,51 +256,99 @@ bool can_parse_error_status(uint32_t err, struct gs_host_frame *frame)
...
+	uint8_t last_tx_error_cnt = (last_err>>16) & 0xFF;
+	uint8_t last_rx_error_cnt = (last_err>>24) & 0xFF;
+	/* If either error counter transitioned to/from 0. */
+	if ((tx_error_cnt == 0) != (last_tx_error_cnt == 0)) {
+		should_send = true;
+	}
+	if ((rx_error_cnt == 0) != (last_rx_error_cnt == 0)) {
+		should_send = true;
+	}
+	/* If either error counter increased by 15. */
```
fenugrec, line 292: is that +15 delta arbitrary , and should it be #defined somewhere ?

src/can.c
```diff
@@ -246,51 +256,99 @@ bool can_parse_error_status(uint32_t err, struct gs_host_frame *frame)
...
+			frame->data[3] |= CAN_ERR_PROT_LOC_CRC_SEQ;
+			should_send = true;
+			break;
+		default: /* 0=no error, 7=no change */
+			break;
 	}
 
-	return true;
+	CAN_TypeDef *can = hcan->instance;
+	/* Write 7 to LEC so we know if it gets set to the same thing again */
```
fenugrec, line 350: good catch !

fenugrec:
rebase + merged in https://github.com/candle-usb/candleLight_fw/commit/ceeac236c4a015af230ae62bbc13f40d3dc08902 , thanks  !

src/can.c
brian-brt: Yes, it's arbitrary. This if statement doesn't normally do anything, because if the error counters are increasing the other code should catch it before they increase by 15. I considered removing it, but decided to leave it in because it helps understand what's going wrong if something else doesn't do what I think it should. Thoughts on what to call a #define?

brian-brt: Ah, I see, you did it. LGTM

PR #62, opened 2020-12-02
Clean up the USB interrupt interactions

brian-brt:
This might fix candle-usb/candleLight_fw#58. It definitely makes it
so I no longer observe similar behavior with multiple hosts running
gs_usb.


src/usbd_gs_can.c
```diff
@@ -680,6 +692,7 @@ uint8_t USBD_GS_CAN_SendFrame(USBD_HandleTypeDef *pdev, struct gs_host_frame *fr
 	USBD_GS_CAN_HandleTypeDef *hcan = (USBD_GS_CAN_HandleTypeDef*)pdev->pClassData;
 	size_t len = sizeof(struct gs_host_frame);
 
+	int primask = disable_irq();
 	if (!hcan->timestamps_enabled)
```
fenugrec, line 696: does this if() block need to be protected from ints ? unless timestamps_enabled can change in an interrupt context... not sure

brian-brt: Yes, `timestamps_enabled` is set from `USBD_GS_CAN_EP0_RxReady` which is one of the callbacks that's run in the USB interrupt. Same with `pad_pkts_to_max_pkt_size`.

src/usbd_gs_can.c
```diff
@@ -601,30 +605,28 @@ static uint8_t USBD_GS_CAN_DataIn(USBD_HandleTypeDef *pdev, uint8_t epnum) {
...
+
+	queue_push_back(hcan->q_from_host, hcan->from_host_buf);
+	hcan->from_host_buf = queue_pop_front(hcan->q_frame_pool);
+	if (hcan->from_host_buf) {
+		USBD_GS_CAN_PrepareReceive(pdev);
+	} else {
+		// gs_can has no way to drop packets. If we just drop this one, gs_can
+		// will fill up its queue of packets awaiting ACKs and then hang. Instead,
+		// wait to call PrepareReceive until we have a frame to receive into.
+	}
```
fenugrec, line 628: Should we think about maybe reorganizing the queues and pool(s) ? How about something like
- separate pools for dataIn/Out so either direction can't starve the other
- dataIn queue > pool size, so a deluge of error frames can't saturate the queue ?
- "allocating" a frame from the pool (maybe requires a separate pool) for every dataOut frame to guarantee we can fit an ACK somewhere  ?

Could be material for a separate issue, definitely needs some thought.

src/usbd_gs_can.c
```diff
@@ -698,8 +710,10 @@ uint8_t USBD_GS_CAN_SendFrame(USBD_HandleTypeDef *pdev, struct gs_host_frame *fr
 		send_addr = buf;
 		len = sizeof(buf);
 	}
```
fenugrec, line 712: Perhaps part of this (memcpy + memset etc) could be safely moved outside of the critical section ? Would make that section much shorter, but of course at the cost of always doing that memcpy+memset regardless of `hcan->pad_pkts_to_max_pkt_size`. This might not be such a big deal; if it doesn't break anything with padding enabled, it should still work at least as well without...

fenugrec (commented):
Left a few notes, most of them either questions or open discussions.

src/usbd_gs_can.c
```diff
@@ -601,30 +605,28 @@ static uint8_t USBD_GS_CAN_DataIn(USBD_HandleTypeDef *pdev, uint8_t epnum) {
...
-		        hcan->from_host_buf = frame;
-		  
-		        retval = USBD_OK;
-		}
-		else{
-		// Discard current packet from host if we have no place
-		// to put the next one
-		}
+	if (rxlen < (sizeof(struct gs_host_frame)-4)) {
+		// Invalid frame length, just ignore it and receive into the same buffer
```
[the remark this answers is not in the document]
brian-brt: Yes, exactly.

src/usbd_gs_can.c
brian-brt: Currently loopback is handled by taking a buffer from the OUT queue and adding it straight back to the IN queue. I can't come up with a way to do that unless they both use the same pool.

I think all the queues already equal to the pool size (`CAN_QUEUE_SIZE`), so the queues never fill up. That definitely makes the error handling easier (the code knows ahead of time it doesn't have a new buffer, vs having to handle a buffer that never makes it into a queue).

I think the current (after this PR) behavior that prioritizes IN over OUT, applies backpressure to both when things get busy, and always has room for the loopback (because it's the same buffer) is pretty reasonable. For OUT data, it fills up the kernel queue nicely (which the user can set the length of), and then it starts informing the userspace application that the queue is full. For IN, it leaves the frames in the hardware FIFO, and then starts dropping once that fills up.

src/usbd_gs_can.c
brian-brt: Good point, done.

src/usbd_gs_can.c
```diff
@@ -601,30 +605,28 @@ static uint8_t USBD_GS_CAN_DataIn(USBD_HandleTypeDef *pdev, uint8_t epnum) {
...
+		// again next time.
+		USBD_GS_CAN_PrepareReceive(pdev);
+		return USBD_OK;
 	}
-	USBD_GS_CAN_PrepareReceive(pdev);
-    return retval;
+
+	queue_push_back(hcan->q_from_host, hcan->from_host_buf);
+	hcan->from_host_buf = queue_pop_front(hcan->q_frame_pool);
```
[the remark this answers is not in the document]
brian-brt: I'm pretty sure Cortex-M0+ does nested interrupts, and interrupts are enabled at this point, which means the queue code could be entered concurrently using the _i versions. I also think this might be the only interrupt being used, which makes it a moot point either way.

Old behavior was to drop the "current" frame (the one which we just finished receiving and are getting the DataOut callback for) if there wasn't a buffer for the "next" one (queue_pop_front from the pool returned NULL), by reusing the buffer and calling USBD_GS_CAN_PrepareReceive again. New behavior is to always put the "current" frame into the queue, and then skip calling USBD_GS_CAN_PrepareReceive if queue_pop_front returns NULL.

Added a comment.

src/usbd_gs_can.c
fenugrec: I guess there's an edge case if the host changes timestamps_enabled or  pad_pkts_to_max_pkt_size without stopping the CAN periph - there could be >=1 more frames sent with the "wrong" ... But that's another story

src/usbd_gs_can.c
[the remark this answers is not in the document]
fenugrec: Thanks for the clarification.

> nested interrupts

ah yes, true.

fenugrec:
Just tested with this PR... I thought all was good, ran a few tests, then left it running a few minutes. Came back, shorted the bus for 5-6 seconds... whole system completely froze, had to reboot... Nothing in the logs of course - I wasn't watching dmesg in realtime and the last errors (if any) didn't make it to /var/log.

Tried to reproduce, didn't manage yet but dmesg has some weird stuff : 
```
Feb 15 19:11:24 raisin kernel: can: controller area network core (rev 20170425 abi 9)
Feb 15 19:11:24 raisin kernel: NET: Registered protocol family 29
Feb 15 19:11:24 raisin kernel: can: raw protocol (rev 20170425)
Feb 15 19:11:57 raisin kernel: gs_usb 1-3:1.0 can1: Unexpected unused echo id 4
Feb 15 19:11:57 raisin kernel: gs_usb 1-3:1.0 can1: Unexpected out of range echo id 62394
Feb 15 19:11:57 raisin kernel: gs_usb 1-3:1.0 can1: Unexpected out of range echo id 1979862773
Feb 15 19:13:11 raisin kernel: gs_usb 5-1:1.0 can0: Unexpected unused echo id 0
```

brian-brt:
The "Unexpected out of range echo id" messages look like memory corruption in your kernel? I've never seen that on either of my test systems. Locking up the system seems like a kernel bug by definition...

I did see some "Unexpected unused echo id" messages before fixing the candleLight side, but never "out of range". I also haven't seen any of those messages using a version with all my fixes (including this one).

For reference, the only bug I've found in the kernel side for gs_usb results in occasional spurious error frames (working on company policy for sending the fix for that upstream).

fenugrec:
Thanks Brian. The more time I spend with this, the less I understand. Supposing the hard lockup was a freak incident, those "out of range echo" are inexplicable. I can reproduce those everytime, but no more system crash.

However, I've managed to get one candle in a Hardfault, as we speak ! No nice stack frame, trying to see what I can recover. I must be lucky or something.

fenugrec:
(WIP, just adding notes as I troubleshoot this)

looks like EP2_OUT has problems : 
```
  OUT_ep = {{
   .............
       }, {                
      num = 0x2,      
      is_in = 0x0,    
      is_stall = 0x0,
      type = 0x2,     
      data_pid_start = 0x0,
      pmaadress = 0x0,
      pmaaddr0 = 0x158,
      pmaaddr1 = 0xd8,                                                         
      doublebuffer = 0x1,
      tx_fifo_num = 0x0,                                                       
      maxpacket = 0x20,           
      xfer_buff = 0x200011c0 "",                                                                                                                               
      xfer_len = 0x0,
      xfer_count = 0x78   
```

 xfer_count = 0x78 ? what ? And xfer_buff points inside a struct now, not a buffer.

Looking at a few lines of PCD_EP_ISR_Handler
```
          if ((ep->xfer_count != 0U) && (ep->xfer_buff != 0U))
          {
            USB_ReadPMA(hpcd->Instance, ep->xfer_buff,
                        ep->pmaadress, (uint16_t)ep->xfer_count);

            ep->xfer_buff += ep->xfer_count;

            /* Process Control Data OUT Packet*/
#if (USE_HAL_PCD_REGISTER_CALLBACKS == 1U)
            hpcd->DataOutStageCallback(hpcd, 0U);
#else
            HAL_PCD_DataOutStageCallback(hpcd, 0U);
#endif /* USE_HAL_PCD_REGISTER_CALLBACKS */
          }
```
Ok, so at this point I decided to check the release notes for more recent versions of the USB stack...


```
1.11.1
HAL/LL USB driver

    Bug fix: USB_ReadPMA() and USB_WritePMA() by ensuring 16-bits access to USB PMA memory
    Bug fix: correct USB RX count calculation
    Fix USB Bulk transfer double buffer mode
    Remove register keyword from USB defined macros as no more supported by C++ compiler
    Minor rework on USBD_Start() and USBD_Stop() APIs: stopping device will be handled by HAL_PCD_DeInit() API.
```
WELL isn't that interesting !!!

fenugrec:
@brian-brt , FYI I rebase this PR onto current master to clean up all the upstream-merges; https://github.com/fenugrec/candleLight_fw/tree/pr62 . But I put that on hold until I try st's USB 1.11.1 update as I feel we're wasting our time with known bugs in 1.11.0

brian-brt:
@fenugrec that works. I'd be happy to rebase too if you want. I just did a merge of master (only conflicts were some whitespace fixes where I removed the code completely). I've found merging preserves github's PR comments/reviews better, but it's kind of a mess no matter what you do so whatever you want.

fenugrec:
No worries -  I was already maintaining that rebase locally anyway, as I find it easier for WIP and tests, as well as for the final merge to master; but you make a good point regarding PR / review readability. My only peference is for at least that one last rebase before merging.

fenugrec:
This patch looks sane, can't find anything objectionable about it, and it doesn't behave worse than the current situation, so merging in 2225fc029008b2837d4309972a80e5787af00bb9 .

Sorry for the delay, these are hard to test and I don't put a lot of time into this ...

PR #63, opened 2020-12-02
Add appropriate memory barriers for disable_irq/enable_irq

brian-brt:
I noticed these were wrong while investigating
candle-usb/candleLight_fw#58. I don't think any of these changes have
any effect right now though, due to the current compiler version and
settings.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -33,7 +33,7 @@ THE SOFTWARE.
 // delay in the effect of enabling and disabling interrupts.
 // That probably doesn't matter here, but it's hard to say what the compiler
 // will put in those 2 instructions so it's safer to leave it. The DSB isn't
-// necessary on Cortex-M0, but it's architecturally required so we'l
+// necessary on Cortex-M0, but it's architecturally required so we'll
 // include it to be safe.
 //
 // The "memory" and "cc" clobbers tell GCC to avoid moving memory loads or
--- a/src/util.c
+++ b/src/util.c
@@ -26,16 +26,34 @@ THE SOFTWARE.
 
 #include <util.h>
 
-inline int disable_irq(void) {
+// ARM's
+// "Application Note 321 ARM Cortex-M Programming Guide to Memory Barrier Instructions"
+// (from https://developer.arm.com/documentation/dai0321/latest) says that
+// the ISBs are actually necessary on Cortex-M0 to avoid a 2-instruction
+// delay in the effect of enabling and disabling interrupts.
+// That probably doesn't matter here, but it's hard to say what the compiler
+// will put in those 2 instructions so it's safer to leave it. The DSB isn't
+// necessary on Cortex-M0, but it's architecturally required so we'l
+// include it to be safe.
+//
+// The "memory" and "cc" clobbers tell GCC to avoid moving memory loads or
+// stores across the instructions. This is important when an interrupt and the
+// code calling disable_irq/enable_irq share memory. The fact that these are
+// non-inlined functions probably forces GCC to flush everything to memory
+// anyways, but trying to outsmart the compiler is a bad strategy (you never
+// know when somebody will turn on LTO or something).
+
+int disable_irq(void) {
 	int primask;
-	asm volatile("mrs %0, PRIMASK\n"
-				"cpsid i\n" : "=r"(primask));
+	asm volatile("dsb sy\nisb sy\nmrs %0, PRIMASK\n"
+				"cpsid i\ndsb sy\nisb sy\n" : "=r"(primask) :: "memory", "cc");
 	return primask & 1;
 }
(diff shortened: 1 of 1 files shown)


fenugrec:
+1 for the explanatory comment in the code.

Any reason not to use the __DSB , __ISB etc primitives from STM32_HAL/include/cmsis/core_gcc.h (includable via core_cmFunc.h) ? interestingly ARM don't seem to ever specify the CC clobber - probably not an issue with their pipeline design ?

fenugrec:
"cc" probably does no harm, but it's probably overkill.
Just a quick look at freeRTOS, for their [CriticalSection](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/master/portable/GCC/ARM_CM0/port.c#L266) they just do
```

 #define portDISABLE_INTERRUPTS()                  __asm volatile ( " cpsid i " ::: "memory" )
 #define portENABLE_INTERRUPTS()                   __asm volatile ( " cpsie i " ::: "memory" )

void vPortEnterCritical( void )
{
    portDISABLE_INTERRUPTS();
    uxCriticalNesting++;
    __asm volatile ( "dsb" ::: "memory" );
    __asm volatile ( "isb" );
}
/*-----------------------------------------------------------*/

void vPortExitCritical( void )
{
    configASSERT( uxCriticalNesting );
    uxCriticalNesting--;

    if( uxCriticalNesting == 0 )
    {
        portENABLE_INTERRUPTS();
    }
}
```
Of course we need to MRS + MSR as well.

brian-brt:
> Any reason not to use the __DSB , __ISB etc primitives from STM32_HAL/include/cmsis/core_gcc.h (includable via core_cmFunc.h) 

None in particular. It was just easier to write it in the inline assembly than figure out how to include that file.

> interestingly ARM don't seem to ever specify the CC clobber - probably not an issue with their pipeline design ?

No, GCC is the only thing that sees the clobbers. It can generate broken assembly code independent of the processor's implementation.

Also, freeRTOS isn't necessarily correct. It tends to have a lot of external functions (like `vPortEnterCritical` and `vPortExitCritical`), which the compiler can't optimize across anyways. Of course, these functions are also external functions, so that applies here too.

I did once get GCC to break code by not including the `"cc"`. However, I don't remember what version it was, and I can't make it happen again. It's possible that was considered a compiler bug and got fixed.

fenugrec:
> No, GCC is the only thing that sees the clobbers. It can generate broken assembly code independent of the processor's implementation.

What I meant is - how can condition codes be affected by an interrupt-related race condition, if their state is atomic within the cpu core ? But,

> I did once get GCC to break code by not including the `"cc"`.

That's interesting, and not too surprising although I don't understand how it could be.

Reminds me of a arm-gcc issue I reported in 2015, still present : C11-style atomics are completely (and silently) broken on M0 (   https://bugs.launchpad.net/gcc-arm-embedded/+bug/1764595 ). So I'm inclined not to trust gcc too much !

brian-brt:
> > No, GCC is the only thing that sees the clobbers. It can generate broken assembly code independent of the processor's implementation.
> 
> What I meant is - how can condition codes be affected by an interrupt-related race condition, if their state is atomic within the cpu core ? But,

GCC took code like this:
```c
asm volatile("cpsid i":::"memory");
if (some_global > 0) {
  other_global = 5;
}
asm volatile("cpsie i":::"memory");
// other stuff, what this is to trigger the bug is the part I forget...
asm volatile("cpsid i":::"memory");
if (some_global > 0) {
  other_global = 3;
}
asm volatile("cpsie i":::"memory");
```

and turned it into something like:
```
cpsid i
ldr r0, [some_global]
cmp r0, #0
ldr r1, =#5
it gt
strgt r1, [other_gloal]
cpsie i
; other stuff
cpsid i
ldr r1, =#3
it gt
strgt r1, [other_global]
cpsie i
```

It's not holding information from memory *in a register* across the inline assembly, it's in the condition code instead.

fenugrec:
Crap, I accidentally pushed master before I was done testing. It was a merge of #27  that also included a different fix for this (see b5cc61b33). That merge had to go in eventually but I'm regretting it now because throughput has decreased and I can't explain why.

brian-brt:
That version using the intrinsics from the CMSIS header looks reasonable to me. If you're good with keeping it, I'd be happy to close this PR.

fenugrec:
I think it looks ok too, albeit less "belt-and-suspenders"... I'll just copy over the comment from this PR , I think it has valuable info re clobbers and ISB  in case we need to look at this again.

fenugrec:
Comment (and an __ISB for good measure) added in 3d52db0a979d2be2da3673792f79e07156ef85a4 . thanks !

PR #66, opened 2021-01-14
Return .user_data section to linker scripts to fix flash saving

tuna-f1sh:
Developing on my [candleLight fork](https://github.com/tuna-f1sh/candleLight_fw) to add support for USB-PD, I needed to add more settings to flash. Doing this caused me to find that the `.user_data` flash sector has been dropped during the linker script migration. 

Basically, the flash saving of user id is not properly defined at the moment and has the potential to over-write application flash.

This PR ports [this commit from 2016](https://github.com/candle-usb/candleLight_fw/commit/1453d70dc9a9d98ac254893ba5114b8e826e0e39), which reduces the application flash region by 2K and allocates a page for flash memory at the end of the flash area. 

I've also added my [new board](https://github.com/tuna-f1sh/entree) as a supported board. I've love to merge all the code from my master fork into this - is there appetite for that? It does [add quite a few things](https://github.com/candle-usb/candleLight_fw/compare/master...tuna-f1sh:master), including i2c but it's all kept board specific so shouldn't impact other hardware.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@ This is firmware for certain STM32F042x/STM32F072xB-based USB-CAN adapters, nota
 - USB2CAN: https://github.com/roboterclubaachen/usb2can (STM32F042x6)
 - CANAlyze: https://kkuchera.github.io/canalyze/ (STM32F042C6)
 - VulCAN Gen1: https://shop.copperforge.cc/products/ac41 (STM32F042x6)
+- Entreé: https://github.com/tuna-f1sh/entree (STM32F042x6)
 
 Of important note is that the common STM32F103 will NOT work with this firmware because its hardware cannot use both USB and CAN simultaneously.
 Beware also the smaller packages in the F042 series which map a USB and CAN_TX signal on the same pin and are therefore unusable !
--- a/ldscripts/STM32F042X6.ld
+++ b/ldscripts/STM32F042X6.ld
@@ -1,10 +1,14 @@
 __STACK_SIZE = 1024;
 __HEAP_SIZE  = 2560;
 
+__FLASH_SIZE = 32K;
+__NVM_SIZE = 1K;
+
 MEMORY
 {
-  FLASH (rx)  : ORIGIN = 0x08000000, LENGTH = 32K
+  FLASH (rx)  : ORIGIN = 0x08000000, LENGTH = __FLASH_SIZE - __NVM_SIZE
   RAM   (rwx) : ORIGIN = 0x20000000, LENGTH = 6K
+  DATA (xrw) : ORIGIN = 0x08000000 + __FLASH_SIZE - __NVM_SIZE, LENGTH = __NVM_SIZE
 }
 
 ENTRY(Reset_Handler)
@@ -68,6 +72,13 @@ SECTIONS
   } > FLASH
 
   __etext = ALIGN (4);
+
+  .user_data :
+  {
+    . = ALIGN(4);
+    *(.user_data)
+    . = ALIGN(4);
+  } > DATA
 
   .data : AT (__etext)
   {
--- a/ldscripts/STM32F072XB.ld
+++ b/ldscripts/STM32F072XB.ld
@@ -1,11 +1,16 @@
 __STACK_SIZE = 2K;
 __HEAP_SIZE  = 10K;
 
+__FLASH_SIZE = 128K;
+__NVM_SIZE = 1K;
+
 MEMORY
 {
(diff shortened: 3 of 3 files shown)


fenugrec:
Whoa, I missed this PR, sorry.

Not sure why that feature was silently dropped  , in https://github.com/candle-usb/candleLight_fw/commit/2ebc665109887bfe95a4ad63ebe5e1d7d13c0fef#diff-579f347859745cf171dd97a898fe970fe0a410e5620b69353887a805eff3c928 ...

It may have something to do with F042 vs F072 memory map differences. Have you compared those ?

Otherwise I agree the current situation can be a problem: as you noticed, .user_data gets linked somewhere in ROM which is clearly not the right thing to do. Just checked a build here and it ends up at 0x080033c8 !

> new board

That depends - 
Extra features applicable to canape only or really out of scope like I2C: sorry, as long as I'm the only one merging stuff, I'll have to decline. If you've looked a bit at the issues and PRs here you'll know I have my plate quite full already trying to test & merge fixes for high-priority, core issues  ...

But if your changes are fixes or clear improvements (like this one), yes, very interested !

tuna-f1sh:
The PR accounts for the F04 and F07 memory size difference. I think it was probably missed during the change to separate linker files. Anyway, I think it should be good to merge?

No problem re new board, probably best as a fork for now. I'll try to help out with some issues and PRs around here regardless since I can merge any fixes into my fork. I use this firmware almost daily so am happy to help.

README.md
```diff
@@ -8,6 +8,7 @@ This is firmware for certain STM32F042x/STM32F072xB-based USB-CAN adapters, nota
 - USB2CAN: https://github.com/roboterclubaachen/usb2can (STM32F042x6)
 - CANAlyze: https://kkuchera.github.io/canalyze/ (STM32F042C6)
 - VulCAN Gen1: https://shop.copperforge.cc/products/ac41 (STM32F042x6)
+- Entreé: https://github.com/tuna-f1sh/entree (STM32F042x6)
```
fenugrec, line 11: I'm happy to add this, but could you split in a different commit ?

ldscripts/STM32F072XB.ld
```diff
@@ -3,8 +3,9 @@ __HEAP_SIZE  = 10K;
 
 MEMORY
 {
-  FLASH (rx)  : ORIGIN = 0x08000000, LENGTH = 128K
+  FLASH (rx)  : ORIGIN = 0x08000000, LENGTH = 128K - 2K
   RAM   (rwx) : ORIGIN = 0x20000000, LENGTH = 16K
+  DATA (xrw) : ORIGIN = 0x801f7ff, LENGTH = 2K
```
fenugrec, line 8: why does DATA start at `0x801f7ff`and not `0x0801F800` (page 63) ?

ldscripts/STM32F042X6.ld
```diff
@@ -3,8 +3,9 @@ __HEAP_SIZE  = 2560;
 
 MEMORY
 {
-  FLASH (rx)  : ORIGIN = 0x08000000, LENGTH = 32K
+  FLASH (rx)  : ORIGIN = 0x08000000, LENGTH = 32K - 2K
   RAM   (rwx) : ORIGIN = 0x20000000, LENGTH = 6K
+  DATA (xrw) : ORIGIN = 0x8007800, LENGTH = 2K
```
fenugrec, line 8: This would mean 2 pages (F042 has 1kB pages), but `flash_flush()` is hardcoded to erase+write 1 page. Might be simpler /safer to limit the size of the DATA area to 1kB on both 042 and 072 ? That _should_ trigger a linker error if anyone tries to pack more than 1k in there

fenugrec (changes requested):

README.md
tuna-f1sh: Done.

ldscripts/STM32F072XB.ld
tuna-f1sh: This was a calculation error on my part - good spot. I've since replace with variables and calculation in file so brought that into this commit instead.

ldscripts/STM32F042X6.ld
tuna-f1sh: Ok, I've reduced to 1k NVM area.

fenugrec (approved):

fenugrec:
Merged in 05fe629e5c47de8b638a9efd1c662c60862a1be9, thanks !

PR #68, opened 2021-02-17
Add basic assert checks

fenugrec:
Avoid using <assert.h> because the standard assert() pulls in
fprintf and its dependencies, as well as including ` __FILE__:__LINE__`
strings for each assert() call. All this has an unacceptable memory
cost, as well as being useless since there is nowhere to printf *to*.

Currently assert_basic() was added to check calloc() return values;
in case of failure we simply halt the core with BKPT(0). This could be
improved and refined.

@brian-brt , @normaldotcom , thoughts / comments ?

### DIFF
--- a/include/util.h
+++ b/include/util.h
@@ -54,7 +54,7 @@ static inline bool is_irq_enabled(void)
 	return (__get_PRIMASK() & 1u) == 0u; // interrupts not prevented
 }
 
-static inline bool disable_irq(void) 
+static inline bool disable_irq(void)
 {
 	bool was_enabled = is_irq_enabled();
 	__disable_irq();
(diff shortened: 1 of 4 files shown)


include/util.h
```diff
@@ -54,17 +54,36 @@ static inline bool is_irq_enabled(void)
...
+ *
+ * Standard assert() needs fprint, __FILE__ , __LINE__ string consts etc.
+ *
+ * stm32's library has "USE_FULL_ASSERT" to enable some checks, but they have
+ * their own assert_param() macro that also passes in __FILE__, __LINE__ so
+ * I'm not sure if (and how) it'll be possible to combine both.
+ *
+ */
+#define assert_basic(exp)         \
+		if (!(exp)) {              \
```
brian-brt, line 81: Wrap this in `do { ... } while (0)` so it's a syntax error to use it without a `;` after it?

brian-brt (commented):
Seems like a good idea to me. I've used `__builtin_trap()` in the past, but it doesn't really matter how you trap into the debugger so this looks fine too.

Just the one comment about making the macro slightly less dangerous.

include/util.h
fenugrec: Oops, the idea was to prevent something like  `if (assert_basic(X)) ...`, but I found the article I was thinking of when I wrote that : https://barrgroup.com/embedded-systems/how-to/define-assert-macro . I left out the last bit; macro should look like
 ```
#define ASSERT(expr) \
    if (expr) \
        {} \
    else \
        aFailed(F_NUM, __LINE__)
```

Will update the PR.

brian-brt (approved):

PR #69, opened 2021-02-23
Berman/add git hash to usb descriptor i product/22 feb21

sgberman:


brian-brt:
@sgberman you've got my changes from #62 in here too. Please remove those.

PR #71, opened 2021-04-12
Add git version string to USB config descriptor

fenugrec:
There was already provisions for this in the code. Tweaked CMake to
generate version.h automatically every build.

This is my proposal for #29 , adding build/version info to USB descriptors.

and under linux, the version string can be seen with lsusb -v  :

```
 iProduct                2 cannette gs_usb
  iSerial                 3 003300264250431420363230
  bNumConfigurations      1
  Configuration Descriptor:
    bLength                 9
    bDescriptorType         2
    wTotalLength       0x0032
    bNumInterfaces          2
    bConfigurationValue     1
    iConfiguration          4 fw_138e6cb_2021-02-17+
    bmAttributes         0x80
      (Bus Powered)
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,6 +28,18 @@ add_subdirectory(libs/STM32_HAL)
 add_subdirectory(libs/STM32_USB_Device_Library)
 set( CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_SOURCE_DIR}/cmake )
 
+
+# Add a custom target that produces version.h, plus
+# a dummy output that's not actually produced, in order
+# to force version.hmake to always be re-run before the build
+
+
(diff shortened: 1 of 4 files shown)


tuna-f1sh:
I just merged it into my fork and tested and it works with no side effects encountered so far 👍 .

Do you think it's worth appending the git hash to the .bin filename in addition? It's easy to track distributed binaries then.

fenugrec:
Thanks for testing ! I'll give this a few more days before merging.

> Do you think it's worth appending the git hash to the .bin filename in addition? It's easy to track distributed binaries then.

Hm, for regular builds I think it could be a pain (i.e. `make clean` might leave traces of older builds when the hash changes?), but it could be something for a CPack target `make package`. We should track that in a separate issue though.

PR #78, opened 2021-12-04
LED : handle 32-bit tick overflow properly

### LINKED ISSUE #77: when 32 bit ms counter folded, led may be ON by mistake? how to avoid that?
https://github.com/candle-usb/candleLight_fw/blob/e6b972441b767b8952d9cd8c78cd74a43be091a9/src/led.c#L94-L98  

49.7 days, after that, folding will happens on an uint32_t counter, ticking every 1ms.
2 ** 32 / 1000 / 60 / 60 / 24 = 49.7 Days

Image this 2 cases:
```
assert( now == 0x0000'FFF3 );
off_until = now + 0x80;
assert( off_until == 0x0001'0073 );
assert( (off_until  < now) == false);

assert( now == 0xFFFF'FFF3 );
off_until = now + 0x80;  // folding happens
assert( off_until == 0x0000'007 );
assert( (off_until  < now) == true);
```

**led may be ON by mistake** when foding happed in case 2.

May this worry realy happens in chance?
Any solution, pattern to solve this folding problem, other than extend 32 bit counter to an endless 64 bit one?

How about:
```
bool is_former_nearly_beyound_by_later(uint32_t former, uint32_t later);
led_set(led, is_former_near_beyound_by_later(led->off_until, now));   
```
"nearly" means diff is not more than half of UINT32_MAX, namely 24.85 days.
```
bool is_former_near_beyound_by_later(uint32_t former, uint32_t later)
{
    uint32_t diff = later - former;
    return (diff & 0x8000'0000) == 0 ? true : false;
}
```
We have cons now: 1) half duration range; 2) slower beyounding detection; however archived "zero mistake chance".

Some experimens shows that folding will not case error any more, after some folding resist way applyed:
https://onlinegdb.com/D2ZBIRjp3 

Cmake scripts, clear code in main loop, led ... candleLight shows an good study example, I learned a lot from this amazing project.
Thanks!

fenugrec:
32-bit milliseconds gives a 49.7-day wrap period which would break the
blinking patterns.

Should fix #77
Tested, with proper logic now.

### DIFF
--- a/src/led.c
+++ b/src/led.c
@@ -28,6 +28,8 @@ THE SOFTWARE.
 #include <string.h>
 #include "hal_include.h"
 
+#define SEQ_ISPASSED(now, target) ((int32_t) ((now) - (target)) >= 0)
+
 void led_init(
 	led_data_t *leds,
 	void* led1_port, uint16_t led1_pin, bool led1_active_high,
@@ -63,8 +65,12 @@ static uint32_t led_set_sequence_step(led_data_t *leds, uint32_t step_num)
 	leds->sequence_step = step_num;
 	led_set(&leds->led_state[0], step->state & 0x01);
 	led_set(&leds->led_state[1], step->state & 0x02);
-	leds->t_sequence_next = HAL_GetTick() + 10*step->time_in_10ms;
-	return 10 * step->time_in_10ms;
+	uint32_t delta = 10 * step->time_in_10ms;
+	if (delta > INT32_MAX) {
+		delta = INT32_MAX;	//clamp
+	}
+	leds->t_sequence_next = HAL_GetTick() + delta;
+	return delta;
 }
 
 void led_run_sequence(led_data_t *leds, led_seq_step_t *sequence, int32_t num_repeat)
@@ -82,7 +88,8 @@ void led_indicate_trx(led_data_t *leds, led_num_t num)
 	uint32_t now = HAL_GetTick();
 	led_state_t *led = &leds->led_state[num];
 
-	if ( (led->on_until < now) && (led->off_until < now) ) {
+	if ( SEQ_ISPASSED(now, led->on_until) &&
+		SEQ_ISPASSED(now, led->off_until) ) {
 		led->off_until = now + 30;
 		led->on_until = now + 45;
 	}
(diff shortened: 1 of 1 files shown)


src/led.c
```diff
@@ -83,7 +87,8 @@ void led_indicate_trx(led_data_t *leds, led_num_t num)
 	uint32_t now = HAL_GetTick();
 	led_state_t *led = &leds->led_state[num];
 
-	if ( (led->on_until < now) && (led->off_until < now) ) {
+	if ( ((int32_t) (led->on_until - now) > 0) &&
```
fenugrec, line 90: uh I think I got this backwards. Will need to look at it again

src/led.c
```diff
@@ -94,7 +99,7 @@ void led_indicate_trx(led_data_t *leds, led_num_t num)
 static void led_update_normal_mode(led_state_t *led)
 {
 	uint32_t now = HAL_GetTick();
-	led_set(led, led->off_until < now);
+	led_set(led, ((int32_t) (led->off_until - now) > 0));
```
fenugrec, line 102: here too

fenugrec:
Thanks for looking at this @daniel-santos .

Not sure I follow your example though : delta is derived from the requested delay, and it cannot be `INT32_MAX + 1` because of the clamping ?
I've changed my mind 3 times while writing this, but currently I think : for the following (UINT32_MAX - 1) ticks after setting the Until value, wouldn't `now - wait_until` still be valid ? e.g. 
```
 now = 0
delta = INT32_MAX (7FFF....)
wait_until = INT32_MAX (7FFF...

// 0 ticks later
(now - wait_until) == -7FFF FFFF ; this is ok

// 1 tick later
(new_now - wait_until) == -7FFF FFFE  ; still ok 

// 7FFF... ticks later
(new_now - wait_until) == 0; ok, now timer tests as "expired"

// new_now == UINT32_MAX - 1 == FFFF FFFE, almost rollover
(new_now - wait_until) == 7FFF FFFF ; ok, timer still tests as "expired"

// new_now == UINT32_MAX == FFFF FFFF,
(new_now - wait_until) == 8000 0000 == -7FFF FFFF; this fails because the sign changed !
```


Maybe I should just clamp to a sane value like 5sec.
In any case, as you say in that other comment you wrote while I was redacting this, I will add macros for this. It's way too error-prone.

daniel-santos:
> Thanks for looking at this @daniel-santos .
> 
> Not sure I follow your example though

That's because I'm a dork and was thinking backwards. :)  Sorry about that, please ignore it!

fenugrec:
> Sorry about that, please ignore it!

No worries, it was a good exercise to go over the whole thing again. This definitely needs to be hidden behind a macro or something, to never be looked at again

PR #80, opened 2022-01-02
Add STM32F4 support

rickyelopez:
Specifically, the STM32F407VE.
Even more specifically, [this](https://stm32-base.org/boards/STM32F407VET6-STM32-F4VE-V2.0.html) dev board


rickyelopez:
@fenugrec I still have to clean this PR up to match the style you've chosen for the rest of the project, but can I get your thoughts? I've tested the basics of this and plan to test more, but is there anything specific you'd like me to look at?

Also, what do you think of including the STM32 HAL libraries as git submodules straight from the STM32 org on github? That way if ever anybody wants to expand this to more boards, they can just add another submodule and the corresponding paths to the CMakeLists?

include/config.h
```diff
@@ -59,7 +69,22 @@ THE SOFTWARE.
 	#define LED2_Pin GPIO_PIN_1
 	#define LED2_Mode GPIO_MODE_OUTPUT_OD
 	#define LED2_Active_High 0
+#elif BOARD == BOARD_STM32F4_DevBoard
+	#define USBD_PRODUCT_STRING_FS		(uint8_t*) "STM32F4VE Dev Board"
+	#define USBD_MANUFACTURER_STRING	(uint8_t*) "idk"
+	#define DFU_INTERFACE_STRING_FS		(uint8_t*) "STM32F4VE firmware upgrade interface"
+	// #define CAN_S_Pin GPIO_PIN_13
```
rickyelopez, line 76: couldn't figure out what this was for so I just disabled it.

include/config.h
```diff
@@ -59,7 +69,22 @@ THE SOFTWARE.
 	#define LED2_Pin GPIO_PIN_1
 	#define LED2_Mode GPIO_MODE_OUTPUT_OD
 	#define LED2_Active_High 0
+#elif BOARD == BOARD_STM32F4_DevBoard
+	#define USBD_PRODUCT_STRING_FS		(uint8_t*) "STM32F4VE Dev Board"
+	#define USBD_MANUFACTURER_STRING	(uint8_t*) "idk"
```
rickyelopez, line 74: update this

ldscripts/STM32F407XE.ld
```diff
@@ -0,0 +1,151 @@
+__STACK_SIZE = 32K;
```
rickyelopez, line 1: not very experienced with linker scripts and didn't know what to choose for stack and heap. This chip has a ton of flash (512k) so this seemed reasonable

ldscripts/STM32F407XE.ld
```diff
@@ -0,0 +1,151 @@
...
+__HEAP_SIZE  = 64K;
+
+__FLASH_SIZE = 512K;
+__NVM_SIZE = 128K;
+
+MEMORY
+{
+  FLASH (rx)  : ORIGIN = 0x08000000, LENGTH = __FLASH_SIZE - __NVM_SIZE
+  RAM   (rwx) : ORIGIN = 0x20000000, LENGTH = 128K
+  DATA (xrw) : ORIGIN = 0x08000000 + __FLASH_SIZE - __NVM_SIZE, LENGTH = __NVM_SIZE
```
rickyelopez, line 11: really unfortunate to have to dedicate a whole 128k to this just to store one uint32, but the last few flash sectors are only addressable in 128k blocks.

libs/STM32_USB_Device_Library/config/usbd_conf.h
```diff
@@ -26,13 +26,21 @@ THE SOFTWARE.
 
 #pragma once
 
+#if defined(STM32F0)
 #include "stm32f0xx_hal.h"
```
rickyelopez: add space here and below

libs/STM32_USB_Device_Library/config/usbd_conf.h
```diff
@@ -26,13 +26,21 @@ THE SOFTWARE.
...
+#define USBD_MAX_NUM_CONFIGURATION    1
+#define USBD_DESC_BUF_SIZE            192
 #define USBD_SUPPORT_USER_STRING_DESC 1
-#define USBD_SELF_POWERED            0
-#define DEVICE_FS                    0
+#define USBD_SELF_POWERED             0
+#if defined(STM32F0)
+# define DEVICE_FS                    0
+#elif defined(STM32F4)
+# define DEVICE_FS                    1
```
rickyelopez: change this back to 0 I think?

src/main.c
```diff
@@ -83,7 +87,7 @@ int main(void)
 	led_set_mode(&hLED, led_mode_off);
 	timer_init();
 
-	can_init(&hCAN, CAN);
+	can_init(&hCAN, CAN_INTERFACE);
```
rickyelopez: this chip actually has two can buses, which I think could both be used without a ton more work. Will look into it.

src/main.c
```diff
@@ -178,42 +182,73 @@ int main(void)
 
 void HAL_MspInit(void)
 {
+    __HAL_RCC_PWR_CLK_ENABLE();
 	__HAL_RCC_SYSCFG_CLK_ENABLE();
+    __HAL_PWR_VOLTAGESCALING_CONFIG(PWR_REGULATOR_VOLTAGE_SCALE1);
```
rickyelopez: This call requires the `pwr` and `pwr_ex` HAL. Not sure if it's actually needed, would be nice to remove.

src/main.c
```diff
@@ -178,42 +182,73 @@ int main(void)
...
 
 void SystemClock_Config(void)
 {
 	RCC_OscInitTypeDef RCC_OscInitStruct;
 	RCC_ClkInitTypeDef RCC_ClkInitStruct;
-	RCC_PeriphCLKInitTypeDef PeriphClkInit;
-	RCC_CRSInitTypeDef RCC_CRSInitStruct;
 
-	RCC_OscInitStruct.OscillatorType = RCC_OSCILLATORTYPE_HSI48;
+#if defined(STM32F4)
```
rickyelopez: delete

src/main.c
```diff
@@ -178,42 +182,73 @@ int main(void)
...
+#endif
+
+	RCC_OscInitStruct.OscillatorType = RCC_OSC_TYPE;
+
+#if defined(STM32F0)
 	RCC_OscInitStruct.HSI48State = RCC_HSI48_ON;
 	RCC_OscInitStruct.PLL.PLLState = RCC_PLL_NONE;
+#elif defined(STM32F4)
+    RCC_OscInitStruct.HSEState = RCC_HSE_ON;
```
rickyelopez: correct indentation

src/main.c
```diff
@@ -178,42 +182,73 @@ int main(void)
...
+#endif
 	HAL_RCC_OscConfig(&RCC_OscInitStruct);
 
-	RCC_ClkInitStruct.ClockType = RCC_CLOCKTYPE_HCLK|RCC_CLOCKTYPE_SYSCLK
-							  |RCC_CLOCKTYPE_PCLK1;
-	RCC_ClkInitStruct.SYSCLKSource = RCC_SYSCLKSOURCE_HSI48;
+	RCC_ClkInitStruct.ClockType = RCC_CLOCKTYPE_HCLK |
+                                  RCC_CLOCKTYPE_SYSCLK |
+#if defined(STM32F4)
+							      RCC_CLOCKTYPE_PCLK1 |
```
rickyelopez: dunno what the heck happened with indentation here...

src/usbd_conf.c
```diff
@@ -32,18 +32,37 @@ PCD_HandleTypeDef hpcd_USB_FS;
...
 {
-	if(hpcd->Instance==USB) {
-		__HAL_RCC_USB_CLK_ENABLE();
-		HAL_NVIC_SetPriority(USB_IRQn, 1, 0);
-		HAL_NVIC_EnableIRQ(USB_IRQn);
+	if(hpcd->Instance==USB_INTERFACE) {
+
+#if defined(STM32F4)
+        __HAL_RCC_GPIOA_CLK_ENABLE();
+        GPIO_InitTypeDef GPIO_InitStruct = {0};
```
rickyelopez: not sure why this isn't necessary for the other chips... pretty sure it's necessary here?

src/usbd_conf.c
```diff
@@ -32,18 +32,37 @@ PCD_HandleTypeDef hpcd_USB_FS;
...
+        GPIO_InitStruct.Mode = GPIO_MODE_AF_PP;
+        GPIO_InitStruct.Pull = GPIO_NOPULL;
+        GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_VERY_HIGH;
+        GPIO_InitStruct.Alternate = GPIO_AF10_OTG_FS;
+        HAL_GPIO_Init(GPIOA, &GPIO_InitStruct);
+#endif
+#if defined(USB)
+	    __HAL_RCC_USB_CLK_ENABLE();
+#elif defined(USB_OTG_FS)
+        __HAL_RCC_USB_OTG_FS_CLK_ENABLE();
```
rickyelopez: fix indentation

src/usbd_conf.c
```diff
@@ -32,18 +32,37 @@ PCD_HandleTypeDef hpcd_USB_FS;
...
 
 void HAL_PCD_MspDeInit(PCD_HandleTypeDef* hpcd)
 {
-	if(hpcd->Instance==USB) {
+	if(hpcd->Instance==USB_INTERFACE) {
+#if defined(USB)
 		__HAL_RCC_USB_CLK_DISABLE();
-		HAL_NVIC_DisableIRQ(USB_IRQn);
+#elif defined(USB_OTG_FS)
+        __HAL_RCC_USB_OTG_FS_CLK_DISABLE();
```
rickyelopez: fix indentation

src/usbd_conf.c
```diff
@@ -105,15 +124,22 @@ USBD_StatusTypeDef USBD_LL_Init(USBD_HandleTypeDef *pdev)
...
-	hpcd_USB_FS.Init.dev_endpoints = 5;
+	hpcd_USB_FS.Instance = USB_INTERFACE;
+	hpcd_USB_FS.Init.dev_endpoints = 5U;
 	hpcd_USB_FS.Init.speed = PCD_SPEED_FULL;
 	hpcd_USB_FS.Init.ep0_mps = EP_MPS_64;
 	hpcd_USB_FS.Init.phy_itface = PCD_PHY_EMBEDDED;
 	hpcd_USB_FS.Init.low_power_enable = DISABLE;
 	hpcd_USB_FS.Init.lpm_enable = DISABLE;
+#if defined(STM32F4)
+    hpcd_USB_FS.Init.dma_enable = DISABLE;
```
rickyelopez: fix indentation

src/usbd_conf.c
```diff
@@ -123,10 +149,20 @@ USBD_StatusTypeDef USBD_LL_Init(USBD_HandleTypeDef *pdev)
...
 	* 0x158 - 0x1D7 (128 bytes) EP1 OUT (buffer 2)
 	*/
-	HAL_PCDEx_PMAConfig((PCD_HandleTypeDef*)pdev->pData , 0x00 , PCD_SNG_BUF, 24);
+#if defined(USB)
+	HAL_PCDEx_PMAConfig((PCD_HandleTypeDef*)pdev->pData , 0x00 , PCD_SNG_BUF, 0x18);
 	HAL_PCDEx_PMAConfig((PCD_HandleTypeDef*)pdev->pData , 0x80 , PCD_SNG_BUF, 0x58);
 	HAL_PCDEx_PMAConfig((PCD_HandleTypeDef*)pdev->pData , 0x81 , PCD_SNG_BUF, 0x98);
 	HAL_PCDEx_PMAConfig((PCD_HandleTypeDef*)pdev->pData , 0x02 , PCD_DBL_BUF, 0x00D80158);
+#elif defined(USB_OTG_FS)
+    // rxFifoSize is in words, so divide number of bytes by 4
```
rickyelopez: remove dead code

src/usbd_gs_can.c
```diff
@@ -271,7 +275,11 @@ static const struct gs_device_bt_const USBD_GS_CAN_btconst = {
 	| GS_CAN_FEATURE_IDENTIFY
 	| GS_CAN_FEATURE_USER_ID
 	| GS_CAN_FEATURE_PAD_PKTS_TO_MAX_PKT_SIZE,
+#if defined(STM32F0)
 	48000000, // can timing base clock
+#elif defined(STM32F4)
+	42000000, // can timing base clock
```
rickyelopez: move this to a `#def` like the other chip-specific constants

normaldotcom:
I took a different (probably way too complicated) approach to adding multitarget support to this firmware with cmake, if you want to reference it see below. Each target has its own linker file, startup file, and HAL conf file. It's a bit of a far diff from the current firmware as I added support for the STM32G431 which has a different CAN peripheral. 

https://github.com/normaldotcom/candleLight_fw/tree/multitarget

Your approach looks far more simple! My only recommendation is to include HAL libraries in a cleaner way to avoid #ifdef's in every c file (ref. `#include TARGET_HAL_LIB_INCLUDE` in my code above, cmake sets this definition).

rickyelopez:
Hey nice that's a pretty nifty way of doing it too! Good point about the includes. I don't think it's all that bad the way I've done it, but could probably write up a pretty easy macro to include the right file. Might take a stab at that at some point!

fenugrec:
Hi, @rickyelopez  I see you responded or are in the process of working on some of your own comments/questions, let me know when this is ready to review.

rickyelopez:
Thansk @fenugrec! Could you please address the following:

> Also, what do you think of including the STM32 HAL libraries as git submodules straight from the STM32 org on github? That way if ever anybody wants to expand this to more boards, they can just add another submodule and the corresponding paths to the CMakeLists?

fenugrec:
> Also, what do you think of including the STM32 HAL libraries as git submodules straight from the STM32 org on github?

I've considered that, but IIRC you can't really get just the HAL on its own, you need to clone the whole CubeF0/4 repo ? Those are huge ; > 300MB just for F0; probably at least as much for F4.
And cloning with --depth 1 doesn't help much - the repos themselves are huge. That annoys me, even more than the current disgusting method of copying the required files locally into this project.

I'm open to suggestions, but I am very averse to cloning an extra 3-4-500MB worth of repos...

rickyelopez:
So you can access just the HAL from their Git:
https://github.com/STMicroelectronics/stm32f0xx_hal_driver
https://github.com/STMicroelectronics/stm32f4xx_hal_driver
but they don't include the CMSIS files, just the HAL drivers (so those would still have to be copied in manually)

I do totally see your concern about file size. These repos are 190Mb each (for shallow clones):
image

Adding them as submodules means that people don't necessarily have to initialize them. You would only have to initialize the one for the chip you want to build, so only an extra 190Mb if you only want to build the F0. Not awful, but fair if you're opposed to it.

include/config.h
fenugrec: Some CAN transceivers have a "Silent mode in which the transmitter is disabled", enabled with this 'S' pin.  Admittedly this should be better documented ; I've just opened #81 to remind myself.

fenugrec:
> So you can access just the HAL from their Git:

Ah thanks, didn't know that. A shallow clone of F0 actually looks like 7MB - that is definitely more acceptable, especially since as you say,

> You would only have to initialize the one for the chip you want to build

For this PR, you seem to have taken care of copying the necessary files , so can we merge like this and discuss submodules separately ? (opened #82 for this)


Also, I like @normaldotcom 's idea to
> include HAL libraries in a cleaner way to avoid #ifdef's in every c file 

This obviously wasn't necessary when everything was F0 but could be worth looking at. Could also be done separately and later. 
(I haven't looked over the whole changeset for this PR)

rickyelopez:
Yeah I was also thinking that could be done in a separate PR, happy to do that.

I've been thinking about how to handle the HAL includes, will take a look at that soon as well for this PR. Thanks for the feedback!

include/config.h
rickyelopez: ohhhh nice yeah that makes sense. Ok I'll add this back in and add a `#def` for it so somebody using this board could hook it up and enable it.

ldscripts/STM32F407XE.ld
fenugrec: IIRC we don't use any heap (at least, not on purpose), so any amount should be ok. Stack usage is pretty minimal - the only difference with F4 will be how their USB-OTG stack is implemented, if they're creating huge buffers on the stack. Unlikely but the right way to check would be to compile with -fstack-usage .

ldscripts/STM32F407XE.ld
fenugrec: oh you mean for the user ID - agreed...  does F4 have other areas where such a value could be stored ? Else, you could always disable the whole user ID section. It's not essential, and in fact was broken for a while...

src/can.c
```diff
@@ -29,326 +29,340 @@ THE SOFTWARE.
...
+      && (sjw>0) && (sjw<=4)
+    ) {
+        hcan->brp = brp & 0x3FF;
+        hcan->phase_seg1 = phase_seg1;
+        hcan->phase_seg2 = phase_seg2;
+        hcan->sjw = sjw;
+        return true;
+    } else {
+        return false;
+    }
```
fenugrec: lots of unrelated whitespace changes in this file. Next time please keep whitespace changes in a different commit to make review easier.

src/can.c
```diff
@@ -29,326 +29,340 @@ THE SOFTWARE.
...
-	itd.Speed = GPIO_SPEED_FREQ_HIGH;
-	itd.Alternate = GPIO_AF4_CAN;
-	HAL_GPIO_Init(GPIOB, &itd);
-
-	hcan->instance   = instance;
-	hcan->brp        = 6;
-	hcan->phase_seg1 = 13;
-	hcan->phase_seg2 = 2;
-	hcan->sjw        = 1;
+    GPIO_InitTypeDef itd;
```
fenugrec: F0 codepath doesn' have  __HAL_RCC_CAN1_CLK_ENABLE  anymore ?

src/flash.c
```diff
@@ -70,13 +74,29 @@ uint32_t flash_get_user_id(uint8_t channel)
...
+#endif
 
 	HAL_FLASH_Unlock();
-	__HAL_FLASH_CLEAR_FLAG(FLASH_FLAG_EOP | FLASH_FLAG_WRPERR | FLASH_SR_PGERR);
-
+	__HAL_FLASH_CLEAR_FLAG(FLASH_FLAG_EOP |
+                           FLASH_FLAG_WRPERR |
+#if defined(STM32F0)
+                           FLASH_SR_PGERR
+#elif defined(STM32F4)
```
fenugrec: this should work but not my favorite method - I usually avoid to #ifdef stuff within a multi-line statement...

src/interrupts.c
```diff
@@ -64,56 +69,143 @@ extern void Reset_Handler();
...
+    0,                    // int 23: I2C1
+    0,                    // int 24: I2C2
+    0,                    // int 25: SPI1
+    0,                    // int 26: SPI2
+    0,                    // int 27: USART1
+    0,                    // int 28: USART2
+    0,                    // int 29: USART3_4
+    0,                    // int 30: CEC_CAN
+    USB_Handler,          // int 31: USB
+#elif defined(STM32F4)
```
fenugrec: hm, ok. At some point we might want to split out stuff that's too different into f0- and f4-specific .c files.

src/main.c
```diff
@@ -61,185 +65,215 @@ uint32_t received_count=0;
...
-	HAL_NVIC_SetPriority(SysTick_IRQn, 0, 0);
+    RCC_OscInitTypeDef RCC_OscInitStruct;
+    RCC_ClkInitTypeDef RCC_ClkInitStruct;
+
+    RCC_OscInitStruct.OscillatorType = RCC_OSC_TYPE;
+
+#if defined(STM32F0)
+    RCC_OscInitStruct.HSI48State = RCC_HSI48_ON;
+    RCC_OscInitStruct.PLL.PLLState = RCC_PLL_NONE;
+#elif defined(STM32F4)
```
fenugrec: I think SystemClock_Config() deserves to be split, what do you think ? that's a lot of #ifdefs !

src/usbd_conf.c
fenugrec: "works here" but I will check better later. aren't we initializing GPIO somewhere else ?

CMakeLists.txt
```diff
@@ -130,18 +130,38 @@ function(add_f072_target TGTNAME)
...
+    target_link_options(STM32_HAL_STM32F072xB BEFORE PRIVATE -mcpu=cortex-m0)
+    target_link_options(STM32_USB_Device_Library_STM32F072xB BEFORE PRIVATE -mcpu=cortex-m0)
     target_link_libraries(${TGTNAME}_fw PRIVATE STM32_HAL_STM32F072xB STM32_USB_Device_Library_STM32F072xB)
     make_bin_file(${TGTNAME}_fw)
     dfu_flash(${TGTNAME}_fw)
     show_object_size(${TGTNAME}_fw)
 endfunction()
 
+function(add_f407_target TGTNAME)
+    set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
```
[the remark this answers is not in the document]
rickyelopez: Yeah I only added it for this target cause it seems like cmake will override the file for each target (so the last target it builds is the only one that gets the file). Since this file is generated in the cmake step, not the make step, it was generating the file for the 072 when I needed it for the 407. 

There's probably a way to get it to generate a file for each target without overwriting them, but it wasn't immediately apparent so I did this.

Ideally, though, yeah this would be enabled for everything.

include/config.h
```diff
@@ -36,12 +36,22 @@ THE SOFTWARE.
...
 #define USBD_INTERFACE_STRING_FS     (uint8_t*) "gs_usb interface"
 
-#define BOARD_candleLight 1
-#define BOARD_cantact     2
-#define BOARD_canable     3
-#define BOARD_usb2can     4
-#define BOARD_canalyze     5
-#define BOARD_cannette    6
+#define BOARD_candleLight      1
+#define BOARD_STM32F4_DevBoard 2
```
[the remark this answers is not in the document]
rickyelopez: originally I had named it "board_candleLight_F4` so I added it under the other candleLight board. When I refactored, I didn't move it. 

I actually think it would be better to just generate a `BOARD_xxxx` for the target that's being generated (rather than `BOARD = BOARD_XXXX`. Then instead of checking if `BOARD == BOARD_xxxx`, you can just check `ifdef BOARD_xxxx`, and you don't need to add all these extra defines.

include/usbd_gs_can.h
```diff
@@ -45,6 +45,21 @@ THE SOFTWARE.
...
+#if defined(STM32F0)
+# define USB_INTERFACE   USB
+# define USB_INTERRUPT   USB_IRQn
+# define CAN_INTERFACE   CAN
+# define CAN_CLOCK_SPEED 48000000
+#elif defined(STM32F4)
+# define USB_INTERFACE   USB_OTG_FS
+# define USB_INTERRUPT   OTG_FS_IRQn
+# define CAN_INTERFACE   CAN1
+# define CAN_CLOCK_SPEED 42000000
```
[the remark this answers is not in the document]
rickyelopez: that's the fastest the peripheral clock can go on this chip. 40MHz would be ideal because then the CAN time quanta can be .15 and it scales up nicely to the typical CAN bitrates, but that's not an option unless we run the sysclk at 40kHz because of the way the dividers are set up.

Since it is technically within spec at 48MHz (and the resulting time quanta), I figured it would be best this way because then we can run the sysclk at 168MHz rather than 40MHz.

include/usbd_gs_can.h
```diff
@@ -45,6 +45,21 @@ THE SOFTWARE.
...
+// RX FIFO is defined in words, so divide bytes by 4
+# define USB_RX_FIFO_SIZE ((24U + 64U + 128U + 128U) / 4U)
```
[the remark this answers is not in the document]
rickyelopez: yup totally. This is actually the formula you have in the `usbd_conf.c` here:
image
I just decided to put it in a `#define` rather than writing out the sum in the function. I can add some text to explain this here, though.

src/main.c
rickyelopez: yep that's totally fair, same for your other comment about the interrupt vector table. I was sort of in the mindset of "let's see how much this is _actually_ different between chips". But yeah, could totally split this up completely and it would probably be simpler.

src/can.c
rickyelopez: Yeah I think vim did this automatically. Are you using tabs or spaces in this project? My vim is set up for spaces which is probably what did this.

ldscripts/STM32F407XE.ld
rickyelopez: yeah I wasn't entirely sure what it was for so I ported it. The first few sectors are much smaller (16k for the first 4, 64k for the next one, and then 128k for the remaining ones). So technically as long as the whole code stays small enough to fit in the first 3 sectors, we can store this in the 4th sector. The rest of the flash becomes unusable at that point though, right? Not super experienced with this sort of thing.

src/usbd_conf.c
rickyelopez: not as far as I could tell, but I definitely could have missed it

src/can.c
rickyelopez: whoops, moved that by accident. will fix, thanks!

ldscripts/STM32F407XE.ld
marckleinebudde: nitpick: stack and heap live in RAM not in ROM (i.e. flash)

rickyelopez: whoops yeah sorry, this chip has 128k of RAM which is where I picked these numbers from, just got confused with flash when writing the comment.

src/usbd_conf.c
fenugrec: src/gpio.c ; we do enable the clocks but we don't explicitly clear GPIO_InitStruct . We probably should

include/config.h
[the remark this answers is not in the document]
fenugrec: > > I actually think it would be better to just generate a `BOARD_xxxx` for the target that's being generated (rather than `BOARD = BOARD_XXXX`

Totally agree. I see no scenario where any code would need the numeric values for those defines.
But for this PR let's keep it simple; just append your board at the end and we can improve this after.

include/usbd_gs_can.h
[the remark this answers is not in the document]
rickyelopez: far as I can tell it works the same in this capacity. This is confirmed working. I'm sure there are nuances and there's probably additional functionality, but this works.

src/usbd_conf.c
rickyelopez: sorry this comment wasn't addressing clearing the struct, I was talking about initializing the USB pins. I couldn't find anywhere where the USB pins are configured for the 042 and 072.

CMakeLists.txt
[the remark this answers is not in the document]
fenugrec: > Yeah I only added it for this target cause it seems like cmake will override the file for each target

Hmm, a quick test here (on master, without this PR), I see all enabled targets in the generated `compile_commands.json` (although after enabling/disabling targets I needed to `make rebuild_cache` to regenerate that file, as expected) ?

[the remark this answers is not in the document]
rickyelopez: hmmmm ok let me try again, maybe I did something wrong when I was testing.

src/can.c
fenugrec: Tabs, as far as I can tell, but possibly not 100% uniform.

rickyelopez: Do you have a preference? If not, I can put up a follow up PR to add clang-format rules and re-format all the files (if you want)

fenugrec: Preference would be : no whitespace change to stuff not directly modified by this PR. clang-format... maybe, not a priority.

rickyelopez: Sorry I meant preference in terms of spaces vs. tabs. I'll revert the changes in this PR.

src/usbd_conf.c
fenugrec: Ah - on those F0, USB isn't an "alternate function" for those pins; they call it "Additional function" ... The RM is not particularly clear about this, but as I understand the GPIO_MODER bits must be kept at "00 - input (reset value").

src/flash.c
rickyelopez: will fix

libs/STM32_HAL/config/hal_include.h
```diff
@@ -0,0 +1,47 @@
...
+// Expand and join tokens
+#define JOIN(x,y)    DO_JOIN(x, y)
+#define JOIN3(x,y,z) JOIN(x, JOIN(y, z))
+
+// Expand and join with `_` separating tokens
+#define SNAKE(x,y) JOIN3(x, _, y)
+
+// macro to include a HAL file for the selected platform
+// (eg. `#include HAL_FILE(hal_gpio)` gives `#include "stm32f4xx_hal_gpio.h"`)
+#define HAL_FILE(file) XSTR(SNAKE(HAL_TARGET_PREFIX, file.h))
```
rickyelopez, line 47: All the defines in this file technically belong in a "utilities.h" or similar, but since these wouldn't be used anywhere other than for this purpose right now, this feels more correct. (if you guys are even OK with this solution)

src/gpio.c
```diff
@@ -90,4 +94,14 @@ void gpio_init()
 	HAL_GPIO_Init(nSI86EN_Port, &GPIO_InitStruct);	//enable si86
 
 #endif // BOARD_cannette
+
+#if BOARD == BOARD_STM32F4_DevBoard
+    // initialize USB pins
+    GPIO_InitStruct.Pin = USB_Pin_DM | USB_Pin_DP;
```
rickyelopez: Added USB init here where it belongs, and only for F4 (from what you wrote in the other comment about the F0 not needing it)

rickyelopez:
Ok I think I addressed all the comments. I came up with a solution for including the right HAL, which involves some macros that could actually be useful for other problems in the future (potentially). Let me know if you guys are ok with this solution, if not I can try to figure something else out. 

This still builds for all the boards, but I don't have any of the other boards to test on. In theory this should be hex-neutral though.

CMakeLists.txt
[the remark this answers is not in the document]
rickyelopez: so yeah there's no overwriting if I define this for the whole file, but it means that vim no longer chooses the right `#def` between STM32F0 and STM32F4, which is pretty annoying when you're trying to develop for one platform specifically.

I moved it out to cover everything for now, but during development people will probably have to move it into the target they're programming for (if their IDEs support reading from this file in the first place).

fenugrec:
> I came up with a solution for including the right HAL, which involves some macros

I saw that, those are indeed nice macros but it seems like overkill at this point. But why not put the `#include HAL_FILE(hal)` inside hal_include.h ? I thought that's what you were going for, .e.g "include hal_include.h to automagically include the right XYZ_hal.h file" .

And at that point, just hiding the
```
#ifdef STM32_F0 
#include <stm32f0xx_hal.h>
#elif ....
```
block inside hal_include.h would be IMO satisfactory (that stm32f0xx_hal.h takes care of including the other _gpio, _can etc. headers), with less preprocessor tricks.


>  In theory this should be hex-neutral though.

Excellent, yes, I just checked before/after and I get the same MD5sum , so you didn't break anything P )

The rest looks generally ok, but I still think the USB RX FIFO size needs to be looked at closer. It's really not equivalent to the PMA layout of the F0 peripheral;  RM0368 (F401) says this in 22.13 :

> Receive FIFO RAM allocation: the application should allocate RAM for SETUP Packets: 10 locations must be reserved in the receive FIFO to receive SETUP packets on control endpoint. The core does not use these locations, which are reserved for SETUP packets, to write any other data. One location is to be allocated for Global OUT NAK. Status information is written to the FIFO along with each received packet. Therefore, a minimum space of (Largest Packet Size / 4) + 1 must be allocated to receive packets. If multiple isochronous endpoints are enabled, then at least two (Largest Packet Size / 4) + 1 spaces must be allocated to receive back-to-back packets. Typically, two (Largest Packet Size / 4) + 1 spaces are recommended so that when the previous packet is being transferred to the CPU, the USB can receive the subsequent packet. Along with the last packet for each endpoint, transfer complete status information is also pushed to the FIFO. Typically, one location for each OUT endpoint is recommended.

rickyelopez:
> But why not put the #include HAL_FILE(hal) inside hal_include.h ?
haha so yeah when I was writing this I had found a few files that included, for example, the `xx_hal_gpio.h` file. Now I realize that's not actually required, you can just include the main `xx_hal.h` file. I'll just go with that (and stash this code somewhere for myself :) )

> so you didn't break anything P )
Nice!

re: PMA
This is the first time I've ever touched this, and tbh, I have no idea what I'm doing. I just tried to port over what you had done for the F0. The text you linked seems to suggest `(Largest Packet Size / 4) + 1` as a good value to use, so the amount of space that I've allocated here is actually way overkill (largest packet is 128k and I've allocated much more than that). Do you think there's any harm to overkill? I'm fine with changing it to `(128k / 4) + 1` and testing that out, can do tonight.

fenugrec:
> > haha so yeah when I was writing this I had found a few files that included, for example, the `xx_hal_gpio.h` file

Fair enough, I also had to double-check that this was in fact unnecessary  !


> Do you think there's any harm to overkill? 

I don't think so, especially since our interaction with USB is designed around the FS peripheral with single-buffered endpoints. 

>  changing it to `(128k / 4) + 1` and testing that out, can do tonight.

128k !? Assuming a typo - I think you can keep the current FIFO sizes you have and replace the PMA comment with something more along the lines of what the RM says.

Can you test your board at heavy bus loads ? maybe a saturated bus at 1MBps or whatever (cangen / canfdtest )-g 0 -p 10

marckleinebudde:
For a max busload test use `cangen` -> `cansequence`:

sender:
`cangen can0 -Di -L0 -I2 -p10 -g0`

receiver:
`cansequence -v can0 -r`

Check the busload via:
`canbusload -t -c -b -r -e can0@1000000`

If this works without problem for an hour, switch sender and receiver.

rickyelopez:
> 128k !? Assuming a typo

Whoops, yeah 128 bytes lol

> replace the PMA comment with something more along the lines of what the RM says.

copy that, can do

> Can you test your board at heavy bus loads

Yup definitely, I'll give it a shot with what @marckleinebudde posted above, and using a PCAN dongle as the receiver (i.e. known good and working at high bus loads)

rickyelopez:
> If this works without problem for an hour, switch sender and receiver.

@marckleinebudde what would I see if anything "doesn't work" in that hour? will errors show up in the terminal that is running one of these commands?

rickyelopez:
Ok so when I have the STM32 as the sender and the pcan dongle as the receiver, I have no issues. Ran it for an hour:
Screenshot from 2022-01-04 20-15-22

But strangely, when I send with the pcan dongle, I can only get up to 50% bus load
Screenshot from 2022-01-04 20-17-29

Not sure if this is an issue with the pcan dongle or the STM32. Can't imagine what would make 
the STM32 have an issue receiving when sending is fine. Any ideas?

Also, I confirmed that upload via DFU works properly!

marckleinebudde:
> > If this works without problem for an hour, switch sender and receiver.
> 
> what would I see if anything "doesn't work" in that hour? will errors show up in the terminal that is running one of these commands?

There can be two problems:

1. The transmission can hang. This results in no `sequence warp around` messages and no bus load.
2. A CAN frame is lost. This results in the receiver showing an error message. With the `-q` option it will quit on error.

You can provoke these errors by stopping the sender (for 1.) and restart the sender (for 2.).

marckleinebudde:
> Not sure if this is an issue with the pcan dongle or the STM32. Can't imagine what would make the STM32 have an issue receiving when sending is fine. Any ideas?

Usually if the receiving node on a CAN network cannot keep up, it will drop CAN frames. There is no flow control on these lower levels on the CAN stack.

So I think it's a limitation of the pcan driver and/or hardware. Do you have another CAN adapter for sending?

fenugrec:
Strange, with two candlelight F0 devices, I'm getting lower busloads when cangen generates short (-L1) frames :

```
$ cangen can1 -Di -L1 -I2 -p10 -g0
 can0@1000000 13486  791415 107888  79%
(cansequence is perfectly happy and reports no dropped frames)

$  cangen can1 -L3 -p10 -g0
 can1@1000000 13450  990643 322800  99%        
```

But, ah... I think this may be USB FSbandwidth limiting us ? we have 64-byte Bulk packets, but it seems the linux gs_usb driver only ever sends one frame per bulk transfer. And there's a hard limit on how many 64-byte transfers will fit per 1ms USB "frame" (as I read table 5-2, USB 2.0 page 42 : procotol overhead (45 bytes), data payload (64) =>  max transfers : 13 per 1ms frame. Very close to the measured 13500, eh....)
Anyway this is for a different discussion. And doesn't explain why your pcan has inferior throughput in the same conditions, or that your F4 board can do 17000 frames per sec...

>> I just checked before/after and I get the same MD5sum , so you didn't break anything P )

Actually I take that back, I wasn't comparing the proper hashes; the F0 builds did change. I suspect `SystemClock_Config` which has some non-trivial changes in the F0 codepath (you moved some stuff around, including __HAL_RCC_CRS_CLK_ENABLE ? , IIRC if you don't enable the clock to a peripheral you can't initialize its registers. I'm surprised it even still works, although there was a bit of weirdness with the LEDs... I suggest you #ifdef the whole (original!) function body instead of trying to segment it. F4 is different enough, IMO there's no point in trying to find common code in low-level init stuff.

marckleinebudde:
> But, ah... I think this may be USB FSbandwidth limiting us? we have 64-byte Bulk packets, but it seems the linux gs_usb driver only ever sends one frame per bulk transfer.

ACK - We can improve the Linux driver to coalesce more CAN frames into a single transfer (Linux -> stm32). Up to the max bulk packet size. Also you can coalesce frames in the stm32 -> Linux direction. Support for this can be negotiated via feature flags.

> And there's a hard limit on how many 64-byte transfers will fit per 1ms USB "frame" (as I read table 5-2, USB 2.0 page 42 : procotol overhead (45 bytes), data payload (64) =>  max transfers : 13 per 1ms frame. Very close to the measured 13500, eh....)

good analysis!

rickyelopez:
> I suspect SystemClock_Config which has some non-trivial changes in the F0 codepath

hmmm that's weird I thought I already reverted those, I'll take another look.

> So I think it's a limitation of the pcan driver and/or hardware. Do you have another CAN adapter for sending?

Very possible. I'm using my work adapter for this test, and we only really go up to 500kbps, dunno if this adapter is very well supported at 1mbps so it sort of makes sense. At 500kbps I can get up to 95% busload no problem:

image

Unfortunately I don't have another adapter for testing :(

Seems like in the context of this PR the only action at this point is for me to remove the hex diff from the F0 boards, right? Are you guys OK with merging without being able to test sending from the F4 at 100% bus load at 1mbps? Or do you have any other proposals for getting that tested?

fenugrec:
> Seems like in the context of this PR the only action at this point is for me to remove the hex diff from the F0 boards, right?

Yes, with the last few changes we discussed (restore SystemClock_Config(),  simplify #include "hal_include.h" mechanism, etc), I'm ok with merging this. High bus load behaviour is IMO a good reason to open a separate issue to improve F4 support.

rickyelopez:
> hmmm that's weird I thought I already reverted those, I'll take another look.

I hadn't pushed :laughing:

fenugrec:
merged  in d3067989  - removed some whitespace changes and reformatted the commit msg, that's why it doesn't match 1:1 with this PR anymore.

Whitespace is a mess in our codebase, I had never really looked closely until now...

Thanks !

rickyelopez:
Nice, thanks! I might put up a PR to address formatting once I get my `uncrustify.cfg` working :)

Issue #84, opened 2022-01-06
CI would be nice. Travis-CI is broken


fenugrec:
Unfortunately the Travis integration added by @vondraussen doesn't work anymore, and now travis-ci is not the best provider for open-source projects.

- github actions : nice integration of course, but there is no out-of-the-box gcc-arm compiler (need to include some build action from an unverified third-party repo...)

- others ?

open to suggestions.

fenugrec:
Initial implementation in  8c5a7d0 . Still need to 

- [ ] update README
- [ ] generate better names for artifacts ?

fenugrec:
Closed via PR #89

PR #86, opened 2022-01-07
add github actions CI #84

vondraussen:
@florolf would this work for you?
the action will install gcc-arm-none-eabi on the ubuntu runner (same as mentioned in [readme](https://github.com/candle-usb/candleLight_fw/blob/master/README.md#building))


.github/workflows/ci.yml
```diff
@@ -0,0 +1,46 @@
...
+    - name: Configure CMake
+      # Configure CMake in a 'build' subdirectory. `CMAKE_BUILD_TYPE` is only required if you are using a single-configuration generator such as make.
+      # See https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html?highlight=cmake_build_type
+      run: cmake -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}} -DCMAKE_TOOLCHAIN_FILE=${{github.workspace}}/cmake/gcc-arm-none-eabi-8-2019-q3-update.cmake
+
+    - name: Build
+      # Build your program with the given configuration
+      run: cmake --build ${{github.workspace}}/build --config ${{env.BUILD_TYPE}}
+
+    - name: Test
```
fenugrec, line 35: We can probably remove this step - there's not many tests we can do in a CI env, with no hardware...

.github/workflows/ci.yml
```diff
@@ -0,0 +1,46 @@
...
+    # The CMake configure and build commands are platform agnostic and should work equally well on Windows or Mac.
+    # You can convert this to a matrix build if you need cross-platform coverage.
+    # See: https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/managing-complex-workflows#using-a-build-matrix
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    
+    - name: Install toolchain
+      run: sudo apt-get -y install gcc-arm-none-eabi
```
fenugrec, line 24: haha... I thought it was going to be more elaborate than this !

.github/workflows/ci.yml
```diff
@@ -0,0 +1,46 @@
...
+      run: cmake --build ${{github.workspace}}/build --config ${{env.BUILD_TYPE}}
+
+    - name: Test
+      working-directory: ${{github.workspace}}/build
+      # Execute tests defined by the CMake configuration.  
+      # See https://cmake.org/cmake/help/latest/manual/ctest.1.html for more detail
+      run: ctest -C ${{env.BUILD_TYPE}}
+      
+    - name: Upload a Build Artifact
+      uses: actions/upload-artifact@v2.3.1
```
fenugrec, line 42: what does `@v2.3.1` mean here ?

vondraussen: that is the version of the **upload-artifact** action. [see here](https://github.com/actions/upload-artifact/tree/v2.3.1)
It allows to store the firmware files, so one can download them after CI run. See on my [fork](https://github.com/vondraussen/candleLight_fw/actions/runs/1667363813), at the bottom you can download **artifact**.
I think it is only stored for 90 days max.

I guess the propper way would be to generate releases based on tags and deploy the artifacts (bin/hex files).

fenugrec: Thanks for clarifying. If we at least get ~ "nightly" builds that will already be a huge improvement. I've been holding off tagging a release for a long time but can revisit this eventually.

vondraussen: Ok, it will build on push or PR to master for now. See definition at the top of the yaml file. Is this okay or do you really want nightly builds?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,34 @@
...
+      run: cmake -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}} -DCMAKE_TOOLCHAIN_FILE=${{github.workspace}}/cmake/gcc-arm-none-eabi-8-2019-q3-update.cmake
+
+    - name: Build
+      run: cmake --build ${{github.workspace}}/build --config ${{env.BUILD_TYPE}}
+
+    - name: Upload a Build Artifact
+      uses: actions/upload-artifact@v2.3.1
+      with:
+        name: firmware
+        path: ${{github.workspace}}/build/*_fw*
```
fenugrec, line 33: Maybe `path: ${{github.workspace}}/build/*_fw*.bin` ? since the ELF file is only really useful for devs.

.github/workflows/ci.yml
fenugrec: > Ok, it will build on push or PR to master for now. See definition at the top of the yaml file. Is this okay 

Yes that is an excellent start. We can always improve or refine it later.

.github/workflows/ci.yml
vondraussen: agree, didn't know they are elfs. 😃

fenugrec:
squashed+merged in 
8c5a7d0 . Thanks for responding so fast !

Issue #87, opened 2022-01-08
firmware echos "hf->flags" which might cause spurious ERRORFRAMEs in Linux


marckleinebudde:
Brian Silverman <brian.silverman@bluerivertech.com> posted the following patch on the Linux CAN mailing list:

https://lore.kernel.org/all/20220106002952.25883-1-brian.silverman@bluerivertech.com/

> No information is deliberately sent here in host->device communications,
> but the open-source candleLight firmware echoes it back, which can
> result in the GS_CAN_FLAG_OVERFLOW flag being set and generating
> spurious ERRORFRAMEs.
> 
> Signed-off-by: Brian Silverman <brian.silverman@bluerivertech.com>
> ```
>  drivers/net/can/usb/gs_usb.c | 1 +
>  1 file changed, 1 insertion(+)
> 
> diff --git a/drivers/net/can/usb/gs_usb.c b/drivers/net/can/usb/gs_usb.c
> index 1b400de00f51..cc4ad8d59bd7 100644
> --- a/drivers/net/can/usb/gs_usb.c
> +++ b/drivers/net/can/usb/gs_usb.c
> @@ -507,6 +507,7 @@ static netdev_tx_t gs_can_start_xmit(struct sk_buff *skb,
> 
>         hf->echo_id = idx;
>         hf->channel = dev->channel;
> +       hf->flags = 0;
> 
>         cf = (struct can_frame *)skb->data;
> ```

marckleinebudde:
FYI: @bsilver8192

PR #88, opened 2022-01-08
main: clear frame->{flags,reserved} of Echo Frames before sending back to host

marckleinebudde:
The Linux driver doesn't clear the flags (and reserved) value of the
struct gs_host_frame send to the device.

With the Echo Frame this value gets send back to Linux und might be
interpreted as a GS_CAN_FLAG_OVERFLOW Error Frame. Clear the flags
value before echoing back. While there clear the reserved value, too.

Since the following patch, Linux clears the flags value before sending it:

| https://lore.kernel.org/all/20220106002952.25883-1-brian.silverman@bluerivertech.com

This patch fixes https://github.com/candle-usb/candleLight_fw/issues/87

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -114,6 +114,8 @@ int main(void)
 		if (frame != 0) { // send can message from host
 			if (can_send(&hCAN, frame)) {
 				// Echo sent frame back to host
+				frame->flags = 0x0;
+				frame->reserved = 0x0;
 				frame->timestamp_us = timer_get();
 				send_to_host_or_enqueue(frame);
 


marckleinebudde:
FYI: @bsilver8192

bsilver8192:
I didn't do this originally because I wasn't sure if any host-side code relies on the current behavior of echoing it back. I know gs_usb doesn't, but something else might. Besides that, LGTM.

fenugrec:
Just catching up now... do I understand correctly that 

1- the `flags` field, from host to device, has no meaning
2- the `flags` field, from device to host, can have meaning (error frames etc)
3- for the echo frames that we send back, we were not clearing `flags` properly ?
4- kernel is going to be "fixed" because of our specific incorrect behaviour (arg... I wish we could just fix these on our end without adding more quirks and edge cases to the kernel !)

bsilver8192:
1. I don't see any place the kernel sets flags being sent host->device (could be hidden in a memcpy somewhere I didn't notice though). Code in this repository definitely doesn't pay any attention to the value there.
2. The kernel has a single bit it checks in the device->host direction (`GS_CAN_FLAG_OVERFLOW`), but nothing in this repository ever sets it. I'm guessing other USB devices using the same driver do.
3. For echo frames, this code was just sending the same value back (echoing it). I'm not sure whether that's proper or not...
4. Leaking uninitialized kernel memory to untrusted USB devices seems undesirable, regardless of whether devices use the values. I think that's a good change to make in the kernel regardless of changes here. It also opens up the possibility to add useful flags there in the future, if devices can trust that kernels which don't support the flags will always send 0.

fenugrec:
Thanks for the comments.

>     4. Leaking uninitialized kernel memory to untrusted USB devices seems undesirable

Fair enough, but doesn't usb_alloc_coherent() clear the buffer ? If not, there must be a lot of other drivers not explicitly clearing other fields before sending stuff over USB...

In any case, this patch seems reasonable.

fenugrec:
rebased + merged in d89d414 . Thanks guys !

PR #89, opened 2022-01-11
add matrix build for separate artifacts #84

vondraussen:
this will add

- cron build (daily)
- build artifacts for each HW target

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,7 @@ jobs:
 
     - name: Build
       env:
-        MAKEFLAGS: "-j4"
+        MAKEFLAGS: "-j"
       run: |
         cmake --build ${{github.workspace}}/build \
               --config ${{env.BUILD_TYPE}} \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,33 +1,53 @@
 name: CI
 
 on:
+  schedule:
+    # build every day at 00:15 UTC
+    - cron:  '15 0 * * *'
   push:
-    branches: [ master ]
   pull_request:
     branches: [ master ]
 
 env:
   BUILD_TYPE: Release
+  CMAKE_TOOLCHAIN_FILE: cmake/gcc-arm-none-eabi-8-2019-q3-update.cmake
 
 jobs:
   build:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        target:
+          - "cantact"
+          - "canalyze"
+          - "canable"
+          - "usb2can"
+          - "cannette"
+          - "candleLight"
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -1,33 +1,53 @@
...
-      run: cmake -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}} -DCMAKE_TOOLCHAIN_FILE=${{github.workspace}}/cmake/gcc-arm-none-eabi-8-2019-q3-update.cmake
+      run: |
+        cmake -B ${{github.workspace}}/build \
+              -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}} \
+              -DCMAKE_TOOLCHAIN_FILE=${{github.workspace}}/${{env.CMAKE_TOOLCHAIN_FILE}}
 
     - name: Build
-      run: cmake --build ${{github.workspace}}/build --config ${{env.BUILD_TYPE}}
+      env:
+        MAKEFLAGS: "-j4"
```
fenugrec, line 43: I was thinking about that the other day, how does the github action VM behave vs -jX settings ? I would've thought they cap all jobs to a certain total %CPU regardless of cores used (to prevent mining and all that crap).

vondraussen: The GitHub runner has 2 virtual cores. Without the option `-jN` only one make process will run. By setting it to four, make will at least run two processes. I just saw, that `-j` should be better, as it would run as many processes as possible.

```
-j [jobs], --jobs[=jobs]Specifies the number of jobs (commands) to run simultaneously. If there is more than one -j option, the last one is effective. If the -j option is given without an argument, make will not limit the number of jobs that can run simultaneously.
```

fenugrec: -j without arguments is usually not a good idea on a real machine, RAM usage can explode (well maybe not on this type of project with just C files, but g++ and Qt can be quite memory hungry).
But, reading
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

it seems we get 2 cores and 7GB RAM (!), so this probably doesn't matter.

vondraussen: Okay, I'll change it.

Shall I add a link to the readme? Right now, one would need to know that binaries can be found under **actions** tab.
Without releases it is not visible to a visitor when just opening the repo/page.

fenugrec: > Shall I add a link to the readme?

Yes, that would be a good idea. I haven't tagged any releases yet, I was hoping to fix #46 first...

fenugrec:
Rebased+merged in 1149642d7 , thanks !

PR #93, opened 2022-03-10
CI : Build targets together and rename artifact

fenugrec:
The "install toolchain" step is very slow, and needed to run for each target part of the build matrix. There is no obvious (to me) way to "cache" the runner environment just before the compilation step.
Building + reusing a docker container seemed to me like a lot of hassle...
Uploading artifacts separately may be possible but I couldn't find how to do that with a build matrix or arrays/maps...

Also, the artifacts are now name differently for Weekly builds, or push/ pullrequest triggered builds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,16 +15,6 @@ env:
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


fenugrec:
Merged

PR #94, opened 2022-03-26
Bugfix#46/fix can hang on host reboot

smalik007:
After running a debugging session with I found two reasons for the can bus going to non responsive state when host reboots.

1. After the host reboot, the `USBD_GS_CAN_Start` function is called, In that we take a pointer out of frame pool and never pushes it back, So every time the host reboot, we leak a pointer one by one, also the `hcan` is not even utilised after taking out the pointer, looks like some old code that has not been cleaned up.
2. The second reason is hard to predict and that's why the occurrence of the issue was so random, It's because the resource lock on `hcan->TxState`, when host reboots, it's not guaranteed that the last msg sent will be delivered and the resource will be unlocked, hence unlocking on resume call back seemed to have tackle this problem. 

Now I have tested the code with this PR for about more than 24 hours where my host system reboots every 2 minute and it seems to be working fine. Cheers

### DIFF
--- a/include/usbd_gs_can.h
+++ b/include/usbd_gs_can.h
@@ -65,6 +65,8 @@ extern USBD_ClassTypeDef USBD_GS_CAN;
 
 uint8_t USBD_GS_CAN_Init(USBD_HandleTypeDef *pdev, queue_t *q_frame_pool, queue_t *q_from_host, led_data_t *leds);
 void USBD_GS_CAN_SetChannel(USBD_HandleTypeDef *pdev, uint8_t channel, can_data_t* handle);
+void USBD_GS_CAN_SuspendCallback(USBD_HandleTypeDef  *pdev);
+void USBD_GS_CAN_ResumeCallback(USBD_HandleTypeDef  *pdev);
 bool USBD_GS_CAN_TxReady(USBD_HandleTypeDef *pdev);
 uint8_t USBD_GS_CAN_PrepareReceive(USBD_HandleTypeDef *pdev);
 bool USBD_GS_CAN_CustomDeviceRequest(USBD_HandleTypeDef *pdev, USBD_SetupReqTypedef *req);
--- a/src/main.c
+++ b/src/main.c
@@ -48,13 +48,13 @@ void SystemClock_Config(void);
 static bool send_to_host_or_enqueue(struct gs_host_frame *frame);
 static void send_to_host();
 
-can_data_t hCAN;
-USBD_HandleTypeDef hUSB;
-led_data_t hLED;
+can_data_t hCAN = {0};
+USBD_HandleTypeDef hUSB = {0};
+led_data_t hLED = {0};
 
-queue_t *q_frame_pool;
-queue_t *q_from_host;
-queue_t *q_to_host;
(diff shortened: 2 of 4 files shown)


fenugrec:
This looks very interesting. Thanks for digging into this issue !

@xdaco , @KeithBoden, @GaryWSmith would you guys have time to test this patch ?

xdaco:
@fenugrec  Hi there
@smalik007  is my colleague. And we are already using the fix in our production code. 

But I would request others to review this fix and get back if you have a better idea.

KeithBoden:
@fenugrec @smalik007 @xdaco My setup was 100% repeatable for the non-responsive state - every reboot would cause the device to be non-responsive and every cold boot would cause the device to function perfectly. I verified this again today on https://github.com/smalik007/candleLight_fw/tree/master and then followed up with the same tests on https://github.com/smalik007/candleLight_fw/tree/bugfix%2346/fix-can-hang-on-host-reboot. The bugfix branch works correctly for me now on reboots, success and thanks!!

fenugrec:
Sorry for the delay. I rebased this in a01c508c86 and took the liberty of rewording some commit messages. +  merged. Thanks again for looking at this annoying issue !

fladdie:
Hello,

I have a Ucan adapter from Fystec and am unfortunately affected by exactly the problem described here. I have flashed the latest firmware. Unfortunately, I still have the problem. If I deactivate the CAN bus interface before the reboot with sudo ip link set down can0, it works again after a reboot. 

I suspect that it has something to do with my setup:
Proxmox as virtualization host and then a USB passthrough to a Linux virtual machine. Maybe the interface does not realize that a reboot is being performed.

Wihmajster:
Sorry for reviving this old thread, but this was the place I stumbled upon when looking up a similar issue to @fladdie, and I thought it might help others who come across it as well.

I had the exact same issue with the UCAN adapter from Fysetc using the original firmware. However, after uploading the latest firmware ([CI#518](https://github.com/candle-usb/candleLight_fw/actions/runs/10876294733), commit: 4ae1a7c2ad06c27d9d10a52a9e28614bab21e430), everything started working perfectly.

I noticed that the Fysetc wiki points to the Canable website for online updater, but from what I saw, that doesn’t seem to use the latest candelight firmware. I didn’t fully verify this, but you may want to skip that step.

Instead, I’d recommend following the instructions in this repository's README, manually downloading the latest FYSETC_UCAN firmware from the CI, and uploading it using DFU using the fail-safe method explained in the documentation.

If the problem persists, I’d suggest opening a new issue on this repo or the UCAN repository to help track down the bug.

KeithBoden:
I have a Canable device and a FYSETC. Interestingly, the Canable works fine after reboots with [4ae1a7c](https://github.com/candle-usb/candleLight_fw/commit/4ae1a7c2ad06c27d9d10a52a9e28614bab21e430) but the FYSETC does not work on a reboot just like the original #46 . I went back to the bugfix branch, [be2ad70a2](https://github.com/smalik007/candleLight_fw/tree/bugfix%2346/fix-can-hang-on-host-reboot) and the FYSETC device works fine. I'll try to find some time to dig into what the difference might be.

PR #97, opened 2022-07-27
Add CANable-MKS board support

ReXT3D:
Introduce firmware build configuration for Makerbase **CANable-MKS** and **CANable-MKS Pro** boards

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -163,7 +163,7 @@ endfunction()
 ########## generate list of targets.
 # the "_fw" part is appended automatically
 set(TGT042_LIST "cantact" "canalyze" "canable" "usb2can" "cannette")
-set(TGT072_LIST "candleLight")
+set(TGT072_LIST "candleLight" "CANable_MKS")
 set(TGT407_LIST "STM32F4_DevBoard")
 
 
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@ This is firmware for certain STM32F042x/STM32F072xB-based USB-CAN adapters, nota
 - CANAlyze: https://kkuchera.github.io/canalyze/ (STM32F042C6)
 - VulCAN Gen1: https://shop.copperforge.cc/products/ac41 (STM32F042x6)
 - Entreé: https://github.com/tuna-f1sh/entree (STM32F042x6)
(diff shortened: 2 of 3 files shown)


ReXT3D:
This is tested and working great on my CANable-MKS Pro board.

I have encountered one curious issue: I had to swap LED1_Pin and LED2_Pin definitions in order for the correct LEDs to light-up on the board.  I had a (very) quick look through the code as well as the board schematic and I do not understand why this reversal is required.

fenugrec:
> I had to swap LED1_Pin and LED2_Pin definitions in order for the correct LEDs to light-up on the board

Can you elaborate on this ? what schematics did you compare, etc. It seems like the "original" canable runs on a F042, and this variant on F072, so a pinout change is not unexpected...

ReXT3D:
> Can you elaborate on this ? what schematics did you compare, etc. It seems like the "original" canable runs on a F042, and this variant on F072, so a pinout change is not unexpected...

Sure thing.  The MKS CANable & Pro have identical pinouts to the candleLight board, except there is no "silent" pin.  For the sake of completeness, the MKS schematic snapshots are:

image
image

In other words, PA0 drives LED1, which is labelled as the TX LED and PA1 drives LED2, which is labelled as the RX LED.  When using firmware with the default candleLight board LED pin assignments I get the RX LED flashing when CANable-MKS is _transmitting_ on the CAN bus and TX LED flashing when _receiving_ on the CAN BUS:

```
#define LED1_Pin GPIO_PIN_0
#define LED2_Pin GPIO_PIN_1
```

I naturally get the correct function when I swap the pin assignments as in this PR:

```
#define LED1_Pin GPIO_PIN_1
#define LED2_Pin GPIO_PIN_0
```

I had a very superficial look at the code and it seems to be addressing LED1 during RX and LED2 during TX in `main.c`.  What confused me is that the candleLight board has reversed LED pin assignment.  But I just confirmed in the KiCad project for the candleLight board that the nomenclature of the two LEDs is indeed reversed vs. the CANable-MKS board.

All is good.  Thank you.

fenugrec:
thanks. Naming the LEDs "led_1" and "led_2" is completely unhelpful, I'll see about renaming them to something smarter

PR #101, opened 2022-08-24
completely reset the CAN peripheral

jhofstee:
In can_enable, the reset bit is used to reset the CAN peripheral, but that does not reset the CAN-state / error counters. Add the RCC reset to make sure the CAN peripheral is completely reset; including state / error counters.

Note: it is only tested for CAN1, I have no hardware which uses CAN2.

### DIFF
--- a/src/can.c
+++ b/src/can.c
@@ -27,6 +27,30 @@ THE SOFTWARE.
 #include "can.h"
 #include "config.h"
 
+// The STM32F0 only has one CAN interface, define it as CAN1 as
+// well, so it doesn't need to be handled separately.
+#if !defined(CAN1) && defined(CAN)
+#define CAN1 CAN
+#endif
+
+// Completely reset the CAN pheriperal, including bus-state and error counters
+static void rcc_reset(CAN_TypeDef *instance)
+{
+#ifdef CAN1
+	if (instance == CAN1) {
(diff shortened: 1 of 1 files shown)


fenugrec:
That looks a bit extreme, was that to fix any specific issues beyond just the counters ? i.e. would just clearing CAN_ESR not be sufficient ?

jhofstee:
First of all, socketcan requires the CAN interface to be in active-error state after open, see [1]. If you don't do that the state in the gs driver and the candlelight firmware are out of sync, and you get unexpected error frames, state changes etc. Writing to the ESR won't help, only the LEC field is writable, the rest is read-only. The only reported way to make that work is by issuing a ECC reset, see [2]. For what it is worth, the end goal is make restart-ms / bit error reporting function correctly, but that is WIP. If at least the CAN starts in a known state, that is at least a step in the right direction.

[1] https://github.com/torvalds/linux/blob/v6.0-rc2/drivers/net/can/usb/gs_usb.c#L843
[2] https://community.st.com/s/question/0D50X00009XkflgSAB/reset-error-condition-in-can

fenugrec:
Ah. Thanks for the extra details. I took the liberty of adding context to the commit message.
rebased & merged

Issue #91, opened 2022-01-30
  gs_usb_breq: GS_USB_BREQ_GET_USER_ID and GS_USB_BREQ_SET_USER_ID


ryedwards:
I am working on a port of candleLight for the STM32G4 and am stepping through updating the flash_flush() carryover function.  I am slowly able to better understand the interaction between the Linux driver and the firmware.  There are two members of the gs_usb_breq, GS_USB_BREQ_GET_USER_ID and GS_USB_BREQ_SET_USER_ID, that are the reason for having the flash functions.  I do not see these being used in either the latest Linux drivers or in cangaroo.  Am I missing something?

The biggest issue I have is that the current code uses a flash page erase to write back a single word.  Since the .user_data section is only defined as a word a page erase will erase a lot more than just that variable.  Does the current linker reserve a whole page (1K on the STM32F04x)?  

There are routines that read an entire page into RAM, erase flash, modify data then write flash (to emulate EEPROM) but I don't see that being done.

Just curious as to the background of the routines in flash.c and how they are intended to function.

fenugrec:
>  current code uses a flash page erase to write back a single word.

Agreed, it's a waste - and it gets worse on bigger devices. The F407 has 128kB pages !

> Does the current linker reserve a whole page 

Yes, see in `ldscripts/ ` , e.g.
```
__NVM_SIZE = 1K;
...
MEMORY
{
  FLASH (rx)  : ORIGIN = 0x08000000, LENGTH = __FLASH_SIZE - __NVM_SIZE
  RAM   (rwx) : ORIGIN = 0x20000000, LENGTH = 6K
  DATA (xrw) : ORIGIN = 0x08000000 + __FLASH_SIZE - __NVM_SIZE, LENGTH = __NVM_SIZE
}

...
```

The only section that should ever end up in DATA, as you found, is `.user_data` . 

I believe this functionality is to emulate behaviour of the original GS_USB hardware. On F042 / F072 losing just 1kB is (usually) not dramatic. I don't think we can fit on the 16kB flash devices anyway ? Haven't checked recently.

If one were to add a new device with the eeprom emulation safely & cleanly disabled I would have no objections to merging. IMO there are better ways to uniquely identify candlelight devices (see #76)

ryedwards:
Thanks for the feedback.  The page size on the G47x is 2kB so it's not horrible to reuse the current code.  The only main difference is that it programs a DOUBLEWORD vs. a WORD.

[...]

PR #102, opened 2022-08-27
RFC : completely remove flash-as-eeprom emulation for get/set_USER_ID

fenugrec:
A follow-up to the discussion in https://github.com/candle-usb/candleLight_fw/issues/91 .

Any thoughts ? There's no rush to delete probably-good code, but if it's not used by anyone...

This would remove all the relevant code and adjusts the linker scripts accordingly.


Daniel-Trevitz:
I've yet to see the value in it. totally unsupported in linux as far as I can tell.

fenugrec:
@HubertD  , any thoughts on this ?

marckleinebudde:
Go ahead, remove that code.

fenugrec:
> The firmware provides proper device unique serial numbers via the USB interface directly.

Yes, I'm just not sure why Hubert went through the trouble of adding a settable user_id ...

Eh, I can always revert the removal if something pops up. But I think it won't be missed.

fenugrec:
Rebase + merged in a0caf93e50edd081db3c6fbc3f36bc720c780a76 .

Issue #103, opened 2022-08-27
we should have `-Wstrict-prototypes`


fenugrec:
and/or enable an optional cppcheck  pass at compile-time

ReXT3D:
FWIW, the current codebase produces only one warning (per board) with `-Wstrict-prototypes` enabled:

```
/home/pi/candleLight_fw/src/interrupts.c:63:1: warning: function declaration isn't a prototype [-Wstrict-prototypes]
 typedef void(*pFunc)();
 ^~~~~~~
```

fenugrec:
Closed via PR #105 . Thanks !

PR #105, opened 2022-08-28
add -Wstrict-prototypes

marckleinebudde:
This PR targets issue: https://github.com/candle-usb/candleLight_fw/issues/103


fenugrec:
Awesome, thanks !

PR #107, opened 2022-09-01
remove odd err-frames

jhofstee:
The gusb-interface can report completely weird error-frames, like being in error-active (based on counters), and error-warning, and bus-off at the same time.
    
  To fix this:
    - only send state changes once.
    - don't send LEC / err counters in bus-off (invalid anyway).
    - don't send empty error frames.
    - remove triggering error message based on counter values / err counters in bus-off (invalid anyway).
    - don't send empty error frames.

### DIFF
--- a/src/can.c
+++ b/src/can.c
@@ -272,7 +272,13 @@ bool can_send(can_data_t *hcan, struct gs_host_frame *frame)
 uint32_t can_get_error_status(can_data_t *hcan)
 {
 	CAN_TypeDef *can = hcan->instance;
-	return can->ESR;
+
+	uint32_t err = can->ESR;
+
+	/* Write 7 to LEC so we know if it gets set to the same thing again */
+	can->ESR = 7<<4;
+
+	return err;
 }
 
 static bool status_is_active(uint32_t err)
@@ -286,6 +292,7 @@ bool can_parse_error_status(uint32_t err, uint32_t last_err, can_data_t *hcan, s
 	 * whether there's anything worth sending. This variable tracks that final
 	 * result. */
 	bool should_send = false;
+	(void) hcan;
 
 	frame->echo_id = 0xFFFFFFFF;
 	frame->can_id  = CAN_ERR_FLAG | CAN_ERR_CRTL;
@@ -322,23 +329,6 @@ bool can_parse_error_status(uint32_t err, uint32_t last_err, can_data_t *hcan, s
 	 * place as any. */
 	frame->data[6] = tx_error_cnt;
 	frame->data[7] = rx_error_cnt;
-
-	uint8_t last_tx_error_cnt = (last_err>>16) & 0xFF;
-	uint8_t last_rx_error_cnt = (last_err>>24) & 0xFF;
-	/* If either error counter transitioned to/from 0. */
-	if ((tx_error_cnt == 0) != (last_tx_error_cnt == 0)) {
-		should_send = true;
-	}
-	if ((rx_error_cnt == 0) != (last_rx_error_cnt == 0)) {
-		should_send = true;
-	}
-	/* If either error counter increased by 15. */
-	if (((int)last_tx_error_cnt + CAN_ERRCOUNT_THRESHOLD) < tx_error_cnt) {
-		should_send = true;
-	}
-	if (((int)last_rx_error_cnt + CAN_ERRCOUNT_THRESHOLD) < rx_error_cnt) {
-		should_send = true;
-	}
 
 	if (err & CAN_ESR_EPVF) {
 		frame->data[1] |= CAN_ERR_CRTL_RX_PASSIVE | CAN_ERR_CRTL_TX_PASSIVE;
@@ -388,10 +378,6 @@ bool can_parse_error_status(uint32_t err, uint32_t last_err, can_data_t *hcan, s
 		default: /* 0=no error, 7=no change */
 			break;
 	}
-
-	CAN_TypeDef *can = hcan->instance;
-	/* Write 7 to LEC so we know if it gets set to the same thing again */
-	can->ESR = 7<<4;
 
 	return should_send;
 }
--- a/src/can.c
+++ b/src/can.c
@@ -295,7 +295,7 @@ bool can_parse_error_status(uint32_t err, uint32_t last_err, can_data_t *hcan, s
 	(void) hcan;
 
 	frame->echo_id = 0xFFFFFFFF;
-	frame->can_id  = CAN_ERR_FLAG | CAN_ERR_CRTL;
+	frame->can_id  = CAN_ERR_FLAG;
 	frame->can_dlc = CAN_ERR_DLC;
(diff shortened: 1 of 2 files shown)


src/can.c
```diff
@@ -270,7 +270,13 @@ bool can_send(can_data_t *hcan, struct gs_host_frame *frame)
 uint32_t can_get_error_status(can_data_t *hcan)
 {
 	CAN_TypeDef *can = hcan->instance;
-	return can->ESR;
+
+	uint32_t err = can->ESR;
+
+	/* Write 7 to LEC so we know if it gets set to the same thing again */
+	can->ESR = 7<<4;
```
fenugrec, line 277: is there a nice macro somewhere we can use instead of a magic 7<<4 constant ? I know this is just original code you moved up here but I just noticed.

src/can.c
```diff
@@ -298,18 +305,23 @@ bool can_parse_error_status(uint32_t err, uint32_t last_err, can_data_t *hcan, s
...
 		}
-	} else if (last_err & CAN_ESR_BOFF) {
-		/* We transitioned out of bus-off. */
-		should_send = true;
+		// - tec (overflowed) / rec (looping, likely used for recessive counting)
+		//   are not valid in the bus-off state.
+		// - The warning flags remains set, error passive will cleared.
+		// - LEC errors will be reported, while the device isn't even allowed to send.
+		//
+		// Hence only report bus-off, ignore everything else.
```
fenugrec, line 318: So transitioning out of bus-off is only signaled via a CAN_ERR_CTRL frame ?

src/can.c
```diff
@@ -298,18 +305,23 @@ bool can_parse_error_status(uint32_t err, uint32_t last_err, can_data_t *hcan, s
...
+		return should_send;
 	}
 
 	/* We transitioned from passive/bus-off to active, so report the edge. */
 	if (!status_is_active(last_err) && status_is_active(err)) {
+		frame->can_id |= CAN_ERR_CRTL;
```
fenugrec, line 324: I just noticed the macro is CAN_ERR_CRTL .... not CTRL ?

src/can.c
```diff
@@ -284,9 +290,10 @@ bool can_parse_error_status(uint32_t err, uint32_t last_err, can_data_t *hcan, s
 	 * whether there's anything worth sending. This variable tracks that final
 	 * result. */
 	bool should_send = false;
+	(void) hcan;
 
 	frame->echo_id = 0xFFFFFFFF;
-	frame->can_id  = CAN_ERR_FLAG | CAN_ERR_CRTL;
+	frame->can_id  = CAN_ERR_FLAG;
```
fenugrec, line 296: note to self : check usage of CAN_ERR_FLAG etc in gs_usb

fenugrec (commented):
Nice, thanks for working on this. Only sending an err frame on changes is indeed more reasonable. I think some of the code you removed with the error counter thresholds is leftover from earlier testing...

src/can.c
fenugrec: Hah, that comes from way upstream... https://github.com/torvalds/linux/blob/master/include/uapi/linux/can/error.h 
I guess we'll learn to say "conrtoller" now

src/can.c
fenugrec: Looks like there is a `CAN_ESR_LEC_Msk`

src/can.c
```diff
@@ -321,35 +333,18 @@ bool can_parse_error_status(uint32_t err, uint32_t last_err, can_data_t *hcan, s
...
 		if (!(last_err & CAN_ESR_EWGF)) {
+			frame->can_id |= CAN_ERR_CRTL;
+			frame->data[1] |= CAN_ERR_CRTL_RX_WARNING | CAN_ERR_CRTL_TX_WARNING;
 			should_send = true;
 		}
-	} else if (last_err & (CAN_ESR_EPVF | CAN_ESR_EWGF)) {
-		should_send = true;
 	}
 
 	uint8_t lec = (err>>4) & 0x07;
```
fenugrec, line 350: Could be done in a separate commit / PR, but this could also use `CAN_ESR_LEC_Pos` / _Msk

src/can.c
fenugrec: What happens if we're still in bus-off (transition happened earlier), but another error sets `should_send = 1`, would we then send an error frame with that new error but with CAN_ERR_BUSOFF now cleared ? Not sure if this can actually happen in hardware but would like to hear your thoughts.

src/can.c
jhofstee: A mask is a mask, not a value, that would be weird to set. An enum might be clearer, but it doesn't seem to exists. can->ESR = (7 << CAN_ESR_LEC_Pos) should work, but since you still need the reference manual to look up the 7, I doubt it improves anything. So to answer your question, no I am not aware there are any nice functions / macros.

src/can.c
jhofstee: as is, CAN_MCR_ABOM is set, it will go from bus-off to error-active, CAN_ERR_CRTL_ACTIVE will be send.

src/can.c
jhofstee: while the peripheral is in bus-off, can_get_error_status will clear the LEC report after the last patch. There shouldn't be any any error to report in bus-off; it is off.

src/can.c
jhofstee: while the peripheral is in bus-off, can_get_error_status will clear the LEC report after the last patch. There shouldn't be any error to report in bus-off; it is off.

src/can.c
jhofstee: I don't mind, this is at least clear. The trouble with mask is things like (value >> shift) & mask vs (value & mask) >> shift. If hard coded it as least clear. And the silicion won't change by magic. For completeness, that is not about code I changed.

src/can.c
fenugrec: nah it's fine. Doesn't need to be looked at for this PR.

fenugrec:
sorry for the delay, haven't had time to look at this and no hardware to test. I'm fairly sure it's ok, but would like either Marc to double-check  if it looks ok w.r.t how the kernel handles error frames, otherwise I will only have time in a few weeks.

jhofstee:
I am not in a hurry, so take your time. For @marckleinebudde I have https://github.com/jhofstee/linux/compare/venus-5.19.4...venus-5.19.4-gsusb as well. Then we can disable / enable the bit-error reporting, access the error-counters and ms-restart should work as it supposed to be (or crash you kernel, it is without locks at the moment e.g. ;)) Well that needs some changes here as well. Anyway, as said, I am not in a hurry, this is the part which doesn't need any changes to the linux driver.

jhofstee:
I guessed it would do that, all whitespace changes.. I will open a new one.

fenugrec:
> I guessed it would do that, all whitespace changes.. I will open a new one.

Sorry about that... it was bound to be a bit messy with outstanding open PRs

Issue #92, opened 2022-02-15
RFC: add support for switchable termination


marckleinebudde:
Hello

@ryedwards and me are planing to add switchable termination to the gs_usb USB interface. The current idea is to add a new feature bit indicating support, and 2 control messages, one to set and one to get the current termination status. Similar to the `IDENTIFY` feature.

I'll add these to the candleLight firmware repo and send a PR when ready.

fenugrec:
Cool. I see mention of something similar in the docs of CANdevstudio, where some drivers support usage such as 
```sudo ip link set can0 type can termination 1```

Is that already in the socketcan layer or other linux drivers, or something entirely driver-specific ?

marckleinebudde:
There is complete support form the user space through the kernel into the CAN drivers. The driver has to implement the `do_set_termination` callback and fill out the `termination_const` and `termination_const_cnt` describing the supported resistor values.

marckleinebudde:
Let's add the interface bits first. See https://github.com/candle-usb/candleLight_fw/pull/109.

PR #108, opened 2022-09-15
Add terminating resistor support

Daniel-Trevitz:
The mcba_usb kernel driver in linux has support for configurable terminating resistors. This patch adds similar support to this firmware which the gs_usb USB driver would then activate via the netlink interface.

I have not tested this code on the device yet or with the linux driver. Currently, I want to know if the maintainers of candleLight are interested in this functionality.

Once I have tested the kernel module changes against these changes I plan to send a similar RFC for the kernel change.

### DIFF
--- a/include/gs_usb.h
+++ b/include/gs_usb.h
@@ -45,6 +45,7 @@ THE SOFTWARE.
 #define GS_CAN_MODE_FD                          (1<<8) /* switch device to CAN-FD mode */
 /* #define GS_CAN_FEATURE_REQ_USB_QUIRK_LPC546XX (1<<9) */
 /* #define GS_CAN_FEATURE_BT_CONST_EXT          (1<<10) */
+/* #define GS_CAN_FEATURE_TERMINATION           (1<<11) */
 
 #define GS_CAN_FEATURE_LISTEN_ONLY              (1<<0)
 #define GS_CAN_FEATURE_LOOP_BACK                (1<<1)
@@ -64,6 +65,12 @@ THE SOFTWARE.
  * GS_USB_BREQ_BT_CONST_EXT and struct gs_device_bt_const_extended
  */
 #define GS_CAN_FEATURE_BT_CONST_EXT             (1<<10)
+/* device supports switchable termination, see:
+ * - GS_USB_BREQ_SET_TERMINATION
+ * - GS_USB_BREQ_GET_TERMINATION
+ * - struct gs_device_termination_state
+ */
+#define GS_CAN_FEATURE_TERMINATION              (1<<11)
 
 #define GS_CAN_FLAG_OVERFLOW                    (1<<0)
 #define GS_CAN_FLAG_FD                          (1<<1) /* is a CAN-FD frame */
@@ -163,6 +170,8 @@ enum gs_usb_breq {
 	GS_USB_BREQ_SET_USER_ID,
 	GS_USB_BREQ_DATA_BITTIMING,
 	GS_USB_BREQ_BT_CONST_EXT,
+	GS_USB_BREQ_SET_TERMINATION,
+	GS_USB_BREQ_GET_TERMINATION,
 };
 
 enum gs_can_mode {
@@ -251,6 +260,10 @@ struct gs_device_bt_const_extended {
 	u32 dbrp_min;
 	u32 dbrp_max;
 	u32 dbrp_inc;
+} __packed;
+
+struct gs_device_termination_state {
+	u32 state;
 } __packed;
 
 struct gs_host_frame {


fenugrec:
(ref - see ticket #92 as well )

Daniel-Trevitz:
@marckleinebudde, Have your efforts on this stalled?

marckleinebudde:
Hello @Daniel-Trevitz,

yes, there was no effort other than the RFC. I'm not sure if @ryedwards build hardware with switchable termination yet. As mentioned in https://github.com/candle-usb/candleLight_fw/issues/92 this should be modeled the same way as the `IDENTIFY` feature. I'll create a patch for the Linux driver implementing these.

Daniel-Trevitz:
[gs_usb.c.txt](https://github.com/candle-usb/candleLight_fw/files/9595078/gs_usb.c.txt)

@marckleinebudde No need, done and works. Just finished testing.

marckleinebudde:
Hello @Daniel-Trevitz,

I've made a pseudo PR to review your driver changes:
https://github.com/marckleinebudde/linux/pull/9#pullrequestreview-1111582359

Overloading the `echo_id` is hack and has no cahnce of going mainline. Here's version of the driver patch:

https://lore.kernel.org/all/20220918202348.675850-1-mkl@pengutronix.de

Daniel-Trevitz:
The revised pull request is not what I am testing with; I have a custom board that uses a neo pixel for status indication.
This is the subset that is relevant to candlelight, with an example added to the dev board for compile testing.

Daniel-Trevitz:
You want me to roll the difference into the PR?

marckleinebudde:
Yes, please rebase ontop of my PR.

Daniel-Trevitz:
Force pushed an update. Still have to rebase off of #109 (learning how to do that...) and actually test it.

src/gpio.c
```diff
@@ -104,3 +118,45 @@ void gpio_init(void)
...
 }
+
+#ifdef PIN_TERM_Pin
+static int term_state = 0;
+#endif
+
+enum terminator_status set_term(unsigned int channel, enum terminator_status state)
+{
+#ifndef PIN_TERM_Pin
+	(void)state;
```
Daniel-Trevitz, line 129: gerrrrr

Daniel-Trevitz:
Testing complete. Changed the dummy implementation in gpio to have fewer ifdefs. I tested by commenting / uncommenting the two new USBD_CtrlError while pretending that the feature was active.

I always get:
sudo ip link set dev can0 type can termination 120
RTNETLINK answers: Broken pipe

include/gpio.h
```diff
@@ -27,3 +27,14 @@ THE SOFTWARE.
 #pragma once
 
 void gpio_init(void);
+
+
+enum terminator_status {
+	term_unsupported = -1,
```
marckleinebudde, line 33: I personally prefer enums and defines in upper case.

include/gpio.h
```diff
@@ -27,3 +27,14 @@ THE SOFTWARE.
 #pragma once
 
 void gpio_init(void);
+
+
```
marckleinebudde, line 31: one newline should be enough

src/gpio.c
```diff
@@ -104,3 +118,50 @@ void gpio_init(void)
 	HAL_GPIO_Init(USB_GPIO_Port, &GPIO_InitStruct);
 #endif
 }
+
+#ifdef PIN_TERM_Pin
+static int term_state = 0;
```
marckleinebudde, line 123: no need to initialize static values to zero.

src/gpio.c
```diff
@@ -104,3 +118,50 @@ void gpio_init(void)
...
+}
+
+enum terminator_status get_term(unsigned int channel)
+{
+	return !!(term_state & (1 << channel));
+}
+
+#else
+
+enum terminator_status set_term(unsigned int channel, enum terminator_status state)
```
marckleinebudde, line 154: Can you place them as `static inline` functions into the header file?

src/usbd_gs_can.c
```diff
@@ -477,10 +489,24 @@ static uint8_t USBD_GS_CAN_Config_Request(USBD_HandleTypeDef *pdev, USBD_SetupRe
 		case GS_USB_BREQ_BITTIMING:
 		case GS_USB_BREQ_IDENTIFY:
 		case GS_USB_BREQ_SET_USER_ID:
+		case GS_USB_BREQ_SET_TERMINATION:
 			hcan->last_setup_request = *req;
```
marckleinebudde, line 493: You can check here with `get_term(req->wValue)` here if that channel supports termination....

src/gpio.c
```diff
@@ -70,6 +71,19 @@ void gpio_init(void)
 	GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_LOW;
 	HAL_GPIO_Init(LEDTX_GPIO_Port, &GPIO_InitStruct);
 
+#ifdef PIN_TERM_Pin
```
marckleinebudde, line 74: If you move the termination gpio init into a separate function, you can have a `static inline` noop or a proper function.

marckleinebudde (changes requested):

marckleinebudde:
> Force pushed an update. Still have to rebase off of #109 (learning how to do that...) and actually test it.

Consider you're on your development branch.

```
# create a backup branch:
git branch backup

# fetch my PR branch into the special branch FETCH_HEAD
git fetch https://github.com/marckleinebudde/candleLight_fw.git termination

# rebase your development branch to FETCH_HEAD
git rebase FETCH_HEAD

# solve conflicts
git add -u
git rebase --continue
```

include/gpio.h
Daniel-Trevitz: I think I'm not going to change that unless it's required. The existing code is tending to lower case.

marckleinebudde: ok

src/usbd_gs_can.c
Daniel-Trevitz: Yeah, that's fair. I had contemplated it, but was tending towards don't do extra work. That said, I think exiting the usb HAL stuff ASAP is better .... and that said I'm probably over thinking this ;)

src/gpio.c
Daniel-Trevitz: ok; I contemplated the same but wanted to follow the existing style.

src/gpio.c
Daniel-Trevitz: This was more or less a choice; specifically calling out zero because an alternative implementation could have a termination on by default.

That and I have a hard time trusting that the data is actually nulled... I mean, I know it is... I just don't trust it.

src/gpio.c
Daniel-Trevitz: Do you mean static inline the version that is a nop? I'm not inclined to expose term_state globally.

Daniel-Trevitz: That does require we pull config.h into gpio.h. Do we care about header files doing that kind of stuff? I can't imagine we do, but might as well check...

src/gpio.c
Daniel-Trevitz: What'd be better is if we had some kind of isolated board support package solution. I know I'd be interested in upstreaming my neo pixel support, but it's just not possible with the way everything is done with ifdefs. I had to add DMA support, interrupts, activate a different GPIO clock, add a whole new c file, etc. etc.

Daniel-Trevitz:
> # rebase your development branch to FETCH_HEAD
> git rebase FETCH_HEAD
Ah ha! I was missing this magic FETCH_HEAD. /thankyou/

Daniel-Trevitz:
I've committed the new version with the rebase; I beleive the static inline of get_term /set_term while in

Daniel-Trevitz:
I've committed the new version with the rebase; I believe the static inline of get_term /set_term while the termination functionality is disabled addresses your underlying concern; not penalizing performance for something that is unused.

It does not fully inline it, since it would move the data to be a public symbol.

src/gpio.c
marckleinebudde: Uninitialized static variables end up in the BSS segment, which is guaranteed by C to be zeroed on startup.

https://stackoverflow.com/questions/9535250/why-is-the-bss-segment-required

include/gpio.h
fenugrec: > The existing code is tending to lower case.

I had to check - it's true, but the sample size is quite small : ) Stuff in gs_usb.h has uppercase enums and is my preference too, but led.c has lowercase. So whatever.

src/gpio.c
```diff
@@ -48,22 +103,17 @@ void gpio_init(void)
 	HAL_GPIO_Init(CAN_S_GPIO_Port, &GPIO_InitStruct);
 #endif
 
-#if (LEDRX_Active_High == 1)
-	HAL_GPIO_WritePin(LEDRX_GPIO_Port, LEDRX_Pin, GPIO_PIN_RESET);
-#else
-	HAL_GPIO_WritePin(LEDRX_GPIO_Port, LEDRX_Pin, GPIO_PIN_SET);
-#endif
+	HAL_GPIO_WritePin(LEDRX_GPIO_Port, LEDRX_Pin, GPIO_INIT_STATE(LEDRX_Active_High));
```
fenugrec, line 106: Ok but this belongs in a separate commit (that would add your GPIO_INIT_STATE macro), as it is not related to terminations

fenugrec (changes requested):
Thanks for working on this.

src/gpio.c
Daniel-Trevitz: Do you want this change? Do you want it in a separate pull request?

src/gpio.c
```diff
@@ -25,8 +25,63 @@ THE SOFTWARE.
...
 #include "hal_include.h"
 
+#define GPIO_INIT_STATE(ACTIVE_HIGH) (ACTIVE_HIGH == 1) ? GPIO_PIN_RESET : GPIO_PIN_SET
+
+#ifdef PIN_TERM_Pin
+static int term_state = 0;
+
+enum terminator_status get_term(unsigned int channel)
+{
+	return !!(term_state & (1 << channel));
```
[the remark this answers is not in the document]
Daniel-Trevitz: I doubt the compiler will remove the conditional, but it'll be interesting to check. I'll make the change at some point here.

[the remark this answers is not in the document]
fenugrec: > I doubt the compiler will remove the conditional

Just checked on godbolt.org and it compiles to same code at -O1 and higher (we use -Os) :

```
enum terminator_status get_term(unsigned int channel)
{
	return !!(term_state & (1 << channel));
}


enum terminator_status get_term2(unsigned int channel)
{
	if (term_state & (1 << channel)) {
		return term_active;
	} else {
		return term_inactive;
	}
}
********************************************
get_term:
        ldr     r3, .L2
        ldr     r3, [r3]
        asrs    r3, r3, r0
        movs    r0, r3
        movs    r3, #1
        ands    r0, r3
        bx      lr
.L2:
        .word   .LANCHOR0
get_term2:
        ldr     r3, .L5
        ldr     r3, [r3]
        asrs    r3, r3, r0
        movs    r0, r3
        movs    r3, #1
        ands    r0, r3
        bx      lr
.L5:
```
Therefore I would prefer if you changed it. Thanks !

(Interestingly, c++ doesn't let you cast bool to enum).

src/gpio.c
fenugrec: Yes please - one small commit for your GPIO_INIT_STATE changes, then the rest of your changes. I just merged #109 so you would need to rebase anyway I think

PR #109, opened 2022-09-19
gs_usb: document switchable termination support

marckleinebudde:
This PR documents the switchable termination support which will be added to the Linux driver soon. For the current Linux driver patch see: https://lore.kernel.org/all/20220918211802.692405-1-mkl@pengutronix.de/

This PR fixes issue https://github.com/candle-usb/candleLight_fw/issues/92

A reference implementation for switchable termination is added to the candleLight firmware in PR https://github.com/candle-usb/candleLight_fw/pull/108.

### DIFF
--- a/include/gs_usb.h
+++ b/include/gs_usb.h
@@ -45,6 +45,7 @@ THE SOFTWARE.
 #define GS_CAN_MODE_FD                          (1<<8) /* switch device to CAN-FD mode */
 /* #define GS_CAN_FEATURE_REQ_USB_QUIRK_LPC546XX (1<<9) */
 /* #define GS_CAN_FEATURE_BT_CONST_EXT          (1<<10) */
+/* #define GS_CAN_FEATURE_TERMINATION           (1<<11) */
 
 #define GS_CAN_FEATURE_LISTEN_ONLY              (1<<0)
 #define GS_CAN_FEATURE_LOOP_BACK                (1<<1)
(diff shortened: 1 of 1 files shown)


marckleinebudde:
Support for switchable termination has been added to the linux-net-next tree and will be mainline with Linux v6.1.

https://git.kernel.org/pub/scm/linux/kernel/git/netdev/net-next.git/commit/?id=906e0e6886afcad6f9cd86660d4b0bdf63f4f200

include/gs_usb.h
```diff
@@ -45,6 +45,7 @@ THE SOFTWARE.
 #define GS_CAN_MODE_FD                          (1<<8) /* switch device to CAN-FD mode */
 /* #define GS_CAN_FEATURE_REQ_USB_QUIRK_LPC546XX (1<<9) */
 /* #define GS_CAN_FEATURE_BT_CONST_EXT          (1<<10) */
+/* #define GS_CAN_FEATURE_TERMINATION           (1<<11) */
```
fenugrec, line 48: This is a weird pattern I hadn't noticed before : this section defines bits for the GS_USB_BREQ_MODE request, shouldn't they all be GS_CAN_MODE_XYZ even when commented out ? Since the next section is for the actual GS_CAN_FEATURE_XYZ flags that specify the supported feature set...

marckleinebudde: Back when I introduced these placeholders my rationale was, there is no `MODE` (e.g. no `GS_CAN_MODE_USB_QUIRK_LPC546XX`) in the Linux driver. It's not supported and never will be. The `MODE` bits are used to communicate from the driver to the firmware on a per port and open base, i.e. the driver enables hardware time stamping on that port while opening it.

On the other hand the `FEATURE` bits communicate from the firmware to the driver that a certain feature is supported (on a per port base). For example the feature `GS_CAN_FEATURE_REQ_USB_QUIRK_LPC546XX` means: Please increase the send USB frame size by 1 byte to work around the USB eratum on LPC546XX. It makes no sense to have a corresponding `MODE` for this. This is why I don't event added a commented out `GS_CAN_MODE_REQ_USB_QUIRK_LPC546XX`.

Hope that makes sense. :smile:

marckleinebudde:
Thanks a lot!

PR #111, opened 2022-10-05
Generalize the gpio configuration

Daniel-Trevitz:
Move the nCANSTBY_Pin to be avaliable for other boards. Make the
  LED Rx and Tx pins optional.

### DIFF
--- a/src/gpio.c
+++ b/src/gpio.c
@@ -27,6 +27,8 @@ THE SOFTWARE.
 #include "config.h"
 #include "hal_include.h"
 
+#define GPIO_INIT_STATE(ACTIVE_HIGH) (((ACTIVE_HIGH) == 1) ? GPIO_PIN_RESET : GPIO_PIN_SET)
+
 // must run before can_init
 void gpio_init(void)
 {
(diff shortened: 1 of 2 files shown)


src/gpio.c
```diff
@@ -27,6 +27,8 @@ THE SOFTWARE.
 #include "config.h"
 #include "hal_include.h"
 
+#define GPIO_INIT_STATE(ACTIVE_HIGH) (ACTIVE_HIGH == 1) ? GPIO_PIN_RESET : GPIO_PIN_SET
```
fenugrec, line 30: Can you add an extra set of () like this : `((ACTIVE_HIGH == 1) ? GPIO_PIN_RESET : GPIO_PIN_SET)`
As a habit, I always enclose the right-hand side in () if it's more complex than a single number.

fenugrec (changes requested):
Looks good, one minor tweak I would like and it's good to go

src/gpio.c
Daniel-Trevitz: you're right. That bit me recently.

fenugrec: oh yes, and it actually needs 2 : one around `(ACTIVE_HIGH)` like you just added, but one around the whole thing would be good too

fenugrec:
Thanks !

Daniel-Trevitz:
@fenugrec, My pleasure! This code has made all kinds of new ideas possible for me!

PR #112, opened 2022-10-06
Use consistent whitespace

Daniel-Trevitz:
Use tabs in cmake to match c file formatting

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,18 +4,18 @@ project(candleLightFirmware C ASM)
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 add_compile_options(
-    -std=gnu11 -mthumb
-    -Wall -Wextra -Werror
-    -Wstrict-prototypes
-    -fmessage-length=0
-    -fsigned-char
-    -ffunction-sections -fdata-sections
-    -ffreestanding
-    -fno-move-loop-invariants
-    -Os -g3
-    -flto -ffat-lto-objects
-    --specs=nano.specs
-    --specs=nosys.specs
+	-std=gnu11 -mthumb
+	-Wall -Wextra -Werror
+	-Wstrict-prototypes
+	-fmessage-length=0
+	-fsigned-char
+	-ffunction-sections -fdata-sections
+	-ffreestanding
+	-fno-move-loop-invariants
+	-Os -g3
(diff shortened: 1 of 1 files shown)


Daniel-Trevitz:
Not sure if you're interested in white space fixes; if not just say so.

marckleinebudde:
LGTM!

fenugrec:
> Not sure if you're interested in white space fixes; if not just say so.

Sure, but if you plan on taking on the entire codebase it would have to be done with some kind of script - either clang-format, uncrustify, astyle, etc. Otherwise it's just a losing battle. Here just for the CMakelists it's fine to do manually, since it's not C

Daniel-Trevitz:
> Sure, but if you plan on taking on the entire codebase it would have to be done with some kind of deterministic script - either clang-format, uncrustify, astyle, etc... Otherwise it's just a losing battle. Here just for the CMakelists it's fine to do manually, since it's not C

Absolutely.

I'm pretty sure there is only one good time to fix whitespace ... before it is written.
Main issue I see is that by making any large whitespace change to the project could create much larger diffs for all of the forks out there (including my own!).

I can look into what it will take to add a code formatter; I think it would be pointless if we don't enforce it with git or github.

Is this something we even want though?

fenugrec:
> Is this something we even want though?

Now and then it will annoy someone - you, sometimes me when I look closely; a while back @rickyelopez was maybe going to start on something : https://github.com/candle-usb/candleLight_fw/pull/80#issuecomment-1006809838

> I think it would be pointless if we don't enforce it with git or github.

Dunno, never used hooks for this. It could perhaps be integrated in the CI checks for each PR ? that would force  the contributor to format their stuff properly at an early stage.

>  all of the forks out there

TBH I never really considered the forks at all - I always hope this repo stays good enough (and generic enough) so that the only forks to appear would be small forks that add, say, a weird uncommon board, or some other minor change that doesn't warrant a PR.

fenugrec:
> It gives me something new to learn, I'll see how much of a PIA it is to do this.

Perhaps this
https://github.com/coleaeason/actions-uncrustify 
could be a starting point. First step would be to at least come up with a sane uncrustify.cfg that minimally changes the codebase - at least with that we can tell contributors how to format their stuff.

fenugrec:
PS - moving the conversation to another ticket

PR #113, opened 2022-10-06
Add switchable terminating resistor support

Daniel-Trevitz:
Add a USB message which controls a GPIO for a specified CAN bus. The GPIO activates a SSR to connect a 120Ohm terminating resistor
  to the specified CAN bus.

### DIFF
--- a/include/config.h
+++ b/include/config.h
@@ -29,8 +29,10 @@ THE SOFTWARE.
 * - LED pin assignments and polarity
 * - other special pins to control CAN transceivers.
 *
-* CAN_S_PIN :Some CAN transceivers (e.g. TJA1050) have a "Silent mode in which the transmitter is disabled";
+* CAN_S_PIN: Some CAN transceivers (e.g. TJA1050) have a "Silent mode in which the transmitter is disabled";
 * enabled with this 'S' pin. If undefined, the corresponding code will be disabled.
+*
+* TERM_Pin: Add support for an externally controlled terminating resistor
 *
 */
 
@@ -206,6 +208,11 @@ THE SOFTWARE.
 	#define USB_GPIO_Port	 GPIOA
 	#define USB_Pin_DM		 GPIO_PIN_11
 	#define USB_Pin_DP		 GPIO_PIN_12
+
+	#define TERM_GPIO_Port		GPIOB
+	#define TERM_Pin			GPIO_PIN_3
+	#define TERM_Mode			GPIO_MODE_OUTPUT_PP
+	#define TERM_Active_High	1
 
 #else
 	#error please define BOARD
--- a/include/gpio.h
+++ b/include/gpio.h
@@ -26,4 +26,27 @@ THE SOFTWARE.
 
 #pragma once
 
+#include "gs_usb.h"
+#include "config.h"
+
 void gpio_init(void);
+
+#ifdef TERM_Pin
+enum gs_can_termination_state set_term(unsigned int channel, enum gs_can_termination_state state);
+enum gs_can_termination_state get_term(unsigned int channel);
+
+#else
+static inline enum gs_can_termination_state set_term(unsigned int channel, enum gs_can_termination_state state)
+{
+	(void)channel;
+	(void)state;
+	return GS_CAN_TERMINATION_UNSUPPORTED;
+}
+
+static inline enum gs_can_termination_state get_term(unsigned int channel)
+{
+	(void)channel;
+	return GS_CAN_TERMINATION_UNSUPPORTED;
+}
+
+#endif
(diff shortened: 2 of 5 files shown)


include/config.h
```diff
@@ -207,6 +209,11 @@ THE SOFTWARE.
 	#define USB_Pin_DM		 GPIO_PIN_11
 	#define USB_Pin_DP		 GPIO_PIN_12
 
+	#define TERM_GPIO_Port		GPIOB
+	#define TERM_Pin			GPIO_PIN_3
+	#define TERM_Mode			GPIO_MODE_OUTPUT_PP
+	#define TERM_Active_High	1
+
```
Daniel-Trevitz, line 216: The indentation of the BOARD_STM32F4_DevBoard section of the config file seems to use a tab size of 4.
github uses a tab size of 8

...I hate tabs...

fenugrec: I used to always use tabsize=4 everywhere and life was good, but now I have mixed feelings, esp. when collaborating with other people / software. Let's ignore the problem for now.

include/gpio.h
```diff
@@ -26,4 +26,32 @@ THE SOFTWARE.
 
 #pragma once
 
+#include "config.h"
+
 void gpio_init(void);
+
+enum terminator_status {
```
fenugrec, line 33: Looking at this a bit closer, that enum is mapped directly to this enum in the kernel gs_usb.c, correct ? :

```
+enum gs_can_termination_state {
+	  = 0,
+	GS_CAN_TERMINATION_STATE_ON
+};
```

( when we handle GS_USB_BREQ_GET_TERMINATION , we're just copying our terminator_status directly in the request)

I'm thinking it may be better to match names (or adjust get_term()) to refer to the same names and values for GS_CAN_TERMINATION_STATE_OFF  etc. ?

Daniel-Trevitz: Well, I’d be fine using the gas usb enum directly. Would defining unsupported as a macro to extend the enum be acceptable?

fenugrec: > Would defining unsupported as a macro to extend the enum be acceptable?

I'm not sure. We're only using that term_unsupported value privately, to determine whether to respond to those USB requests. We could of course #ifdef those requests out, but I find we already have a lot of ifdef's already. The way you have it currently is probably equivalent, since get_term is static inline returning a constant - so long as it gets compiled out.

Did you mean replacing our enum with something like 
```
#define GS_CAN_TERMINATION_UNSUPPORTED -1    // private, not in kernel enum
enum gs_can_termination_state {
	GS_CAN_TERMINATION_STATE_OFF = 0,
	GS_CAN_TERMINATION_STATE_ON,
};
```
? Probably ok too. As long as we keep our enum in sync with the kernel's...

Daniel-Trevitz: Macro made gcc angry. I just put it in the enum so it'd shutup about invalid type comparisons.

src/usbd_gs_can.c
```diff
@@ -383,6 +388,13 @@ static uint8_t USBD_GS_CAN_EP0_RxReady(USBD_HandleTypeDef *pdev) {
 			}
 			break;
 
+		case GS_USB_BREQ_SET_TERMINATION:
+			memcpy(&param_u32, hcan->ep0_buf, sizeof(param_u32));
+			if (set_term(req->wValue, param_u32) == GS_CAN_TERMINATION_UNSUPPORTED) {
+				USBD_CtlError(pdev, req);
```
fenugrec, line 394: Is this the proper way to handle the case where the host sends a SET_TERMINATION request even if GS_CAN_FEATURE_TERMINATION wasn't set ? especially since when handling the request (line 497) we're already aborting with USBD_CtlError()

fenugrec: i.e. maybe this would be sufficient

```
if (get_term(req->wValue) == GS_CAN_TERMINATION_UNSUPPORTED) {
	break;
}
memcpy(&param_u32, hcan->ep0_buf, sizeof(param_u32));
(void) set_term(req->wValue, param_u32);
```

Daniel-Trevitz: Throwing the USB error causes the linux driver to error out which reports back to the user of an issue. I.E. The ip command complains when it's not supported.

fenugrec:
Rebased & merged in 0a1d6dc920bb8c243c7bcae5b54b10f9282e912c , thanks.

With a few concurrent changes (whitespace, some local work on #includes, etc) it was getting messy to keep this PR around so I merged mostly as-is.

I still have reservations regarding the handling of the new _GET and _SET_TERMINATION requests but opened a new ticket for this, #116

PR #115, opened 2022-10-11
Uncrustify the code

Daniel-Trevitz:
Add an uncrustify config file and apply the changes it creates.

### DIFF
--- a/include/can.h
+++ b/include/can.h
@@ -57,7 +57,7 @@ bool can_send(can_data_t *hcan, struct gs_host_frame *frame);
  */
 uint32_t can_get_error_status(can_data_t *hcan);
 
-#define CAN_ERRCOUNT_THRESHOLD 15	/* send an error frame if tx/rx counters increase by more than this amount */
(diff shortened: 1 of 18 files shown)


Daniel-Trevitz:
The config has most of the rules off, only enabling the rules that seem to be most obvious.

marckleinebudde:
<del>Let the bike shedding begin!</del>

What abou using the Linux kernel coding style?

https://github.com/uncrustify/uncrustify/blob/master/etc/linux.cfg

fenugrec:
> ~Let the bike shedding begin!~
> What abou using the Linux kernel coding style?

Because 1TBS is the True Way !

But more seriously, I'm hoping for just a minimal .cfg that takes care of the worst whitespace offences while changing as little as possible - I believe this commit manages fairly well, but will look closer when I have more time.

fenugrec:
Thanks ! 
I've no problem adjusting the .cfg file as we find new inconsistencies. At least we have a starting point now

Issue #116, opened 2022-10-12
Handling GS_USB_BREQ_SET_TERMINATION and GS_USB_BREQ_GET_TERMINATION requests


fenugrec:
Continuing discussion from #113.

There seems to be a difference in behaviour between old firmware, and new firmware compiled without TERM_PIN , when receiving those new requests. 

Originally, in `USBD_GS_CAN_Config_Request()`, any unsupported request falls through to the default, `USBD_CtlError(pdev, req);` this seems to be also the case in the new code but with a few added steps (adds about 30 bytes; gcc isn't quite able to optimize out everything  unless maybe when compiling with `-flto`)

in `USBD_GS_CAN_EP0_RxReady()`, previous code would do nothing and return USBD_OK, but now we  `USBD_CtlError(pdev, req);` .

Daniel-Trevitz:
I suspect that USBD_GS_CAN_EP0_RxReady will never be called as the USBD_GS_CAN_Config_Request drops it out first; I.E. the check there is uneeded.

unless you replace the param_u32 with a constant, the if statement will not be optimized out.
The memcpy can not be optimized out.

This is optimized out:
		case GS_USB_BREQ_SET_TERMINATION:
			set_term(req->wValue, *(uint32_t*)hcan->ep0_buf);
			break;

This is insufficient for setting an invalid terminating resistor when it's enabled.

found it. PR coming in hot

fenugrec:
Closed via pr #117, thanks

PR #117, opened 2022-10-12
Help the compiler

Daniel-Trevitz:
Addresses issue #116


Daniel-Trevitz:
That was a fun rebase... got to watch out for that goof...

fenugrec:
heh, I had exactly that same change in a local branch, you beat me to it P )

I was debating whether or not to #ifdef those blocks but like this is probably more maintenance-free.

Daniel-Trevitz:
minimizing ifdefs is a worthy goal. Only when the other choices are ~10 orders of magnitude worse should it be considered...

PR #119, opened 2022-10-14
Errv2,  remove odd err-frames

jhofstee:
The same as https://github.com/candle-usb/candleLight_fw/pull/107, but after the white space changes:

remove odd err-frames

The gusb-interface can report completely weird error-frames,
like being in error-ative (based on counters), and error-warning,
and bus-off at the same time.

To fix this:
  - only send state changes once.
  - don't send LEC / err counters in bus-off (invalid anyway).
  - don't send empty error frames.

### DIFF
--- a/include/can.h
+++ b/include/can.h
@@ -57,8 +57,6 @@ bool can_send(can_data_t *hcan, struct gs_host_frame *frame);
  */
 uint32_t can_get_error_status(can_data_t *hcan);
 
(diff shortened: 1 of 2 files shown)


jhofstee:
lovely this @marckleinebudde, but at least it is a bit better. Afaic this needs to be rewritten.

fenugrec:
rebased + merged . Thanks !

PR #127, opened 2022-11-09
Prepare for stm32g0

marckleinebudde:
Take the first non STM32G0 specific patches from https://github.com/candle-usb/candleLight_fw/pull/126 and clean them up.


fenugrec:
awesome. I was going to do something similar but have been busy.

fenugrec:
Thanks !

PR #128, opened 2022-11-10
queue: remove unused functionality

marckleinebudde:
Looking at the queue code it turns out, a lot of the functions are not used. Remove the unused queue functionality. Not sure why the compiler hasn't removed the `malloc()` and `free()` functions entirely, but this saves some bytes:

```
| add/remove: 0/3 grow/shrink: 0/0 up/down: 0/-232 (-232)
| Function                                     old     new   delta
| malloc                                        20       -     -20
| free                                          20       -     -20
| _free_r                                      192       -    -192
| Total: Before=17370, After=17138, chg -1.34%
```

### DIFF
--- a/include/queue.h
+++ b/include/queue.h
@@ -36,16 +36,7 @@ typedef struct {
 } queue_t;
 
 queue_t *queue_create(unsigned max_elements);
-void queue_free(queue_t *q);
 
-unsigned queue_size(queue_t *q);
-bool queue_is_empty(queue_t *q);
 bool queue_push_back(queue_t *q, void *el);
 bool queue_push_front(queue_t *q, void *el);
 void *queue_pop_front(queue_t *q);
-
-unsigned queue_size_i(queue_t *q);
-bool queue_is_empty_i(queue_t *q);
-bool queue_push_back_i(queue_t *q, void *el);
-bool queue_push_front_i(queue_t *q, void *el);
-void *queue_pop_front_i(queue_t *q);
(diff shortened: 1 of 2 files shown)


fenugrec:
Interesting, I'm not seeing exactly the same gains (arm-gcc 12.2) ; malloc dissappears but we still have calloc due to `queue_create()` . Here I did not have any users of "free" either.

What is strange to me is without this commit, atexit calls malloc (!!) via `__register_exitproc`. Applying this commit changes the implementation to one without malloc...

what compiler are you using ?

fenugrec:
Also some interesting comments re malloc here https://stackoverflow.com/questions/48711221/how-to-prevent-inclusion-of-c-library-destructors-and-atexit

marckleinebudde:
> Interesting, I'm not seeing exactly the same gains (arm-gcc 12.2) ; malloc dissappears but we still have calloc due to `queue_create()` . Here I did not have any users of "free" either.

I'm currently working on converting the queue implementation to Linux compatible lists. They can allocated on the stack or BSS easier.

> What is strange to me is without this commit, atexit calls malloc (!!) via `__register_exitproc`. Applying this commit changes the implementation to one without malloc...
> 
> what compiler are you using ?

`arm-none-eabi-gcc (15:11.3.rel1-1) 11.3.1 20220712` from Debian testing.

fenugrec:
> I'm currently working on converting the queue implementation to Linux compatible lists. They can allocated on the stack or BSS easier.

Cool stuff.

In any case, this is good to go. Thanks !

fenugrec:
Rebased & merged in 7f973211499ff0c58101765f2880ccc6b8a3465b

PR #129, opened 2022-11-10
Cmake cleanups

marckleinebudde:
First clean up the cmake file a bit, then add the `--print-memory-usage` linker option to print a overview of the memory usage.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,9 +33,9 @@ add_link_options(
 	--specs=nosys.specs
 	-Wall
 	-Wextra
-	-Xlinker --gc-sections
 	-g3
 	-mthumb
+	LINKER:--gc-sections
 )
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -4,29 +4,39 @@ project(candleLightFirmware C ASM)
...
 add_link_options(
-	-mthumb -Wall -Wextra -g3
-	-Xlinker --gc-sections
 	--specs=nano.specs
 	--specs=nosys.specs
+	-Wall
+	-Wextra
+	-g3
+	-mthumb
+	LINKER:--gc-sections
```
fenugrec, line 38: Cool. I thought that was a new cmake feature, but it's already in 3.13 which is what we're requiring already

fenugrec:
thanks !

PR #131, opened 2022-11-11
gs_usb cleanups

marckleinebudde:
First some cleanups, then mark `struct gs_*` as aligned to 4 bytes.

### DIFF
--- a/include/compiler.h
+++ b/include/compiler.h
@@ -1,0 +1,116 @@
+/*-
+ * Copyright (c) 2010 Isilon Systems, Inc.
+ * Copyright (c) 2010 iX Systems, Inc.
+ * Copyright (c) 2010 Panasas, Inc.
+ * Copyright (c) 2013-2016 Mellanox Technologies, Ltd.
+ * Copyright (c) 2015 François Tigeot
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice unmodified, this list of conditions, and the following
+ *    disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE AUTHOR ``AS IS'' AND ANY EXPRESS OR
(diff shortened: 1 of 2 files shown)


fenugrec:
- Do we want some kind of more-compiler-portable method of marking aligned data ? Granted we're pretty much tied to gcc right now.

- we're already requiring C11 support, we could use a `Static_assert()` to validate correct struct size at compile-time ?

marckleinebudde:
> Do we want some kind of more-compiler-portable method of marking aligned data?

These macros are defined in the `STM32_HAL` to make it portable.

> Granted we're pretty much tied to gcc right now.

The only other compiler I can think of is `clang`, we can solve the problem, once it hits us :)

> we're already requiring C11 support, we could use a Static_assert() to validate correct struct size at compile-time ?

Yes could be done, but I don't see the benefit.

Now we say the compiler, don't add holes and assume it's aligned to 4 bytes.

If you have neither packed nor aligned, the compiler assumes 4 bytes alignment. This is how I found the issue, I tried to assign a pointer to a packed struct to a pointer of a non packed struct and the compiler warned that the packed pointer doesn't have a 4 byte alignment.

marckleinebudde:
After seeping over it, decided to use `__aligned` as in the Linux kernel and add a minimal `compiler.h`. The `__ALIGN_BEGIN` and `__ALIGN_END` ist just too ugly :)

marckleinebudde:
changes:
* imported and improved `compiler.h` from FreeBSD
* use `__algined()` instead of `__ALIGN_BEGIN` and `__ALIGN_END`

fenugrec:
>     * imported and improved `compiler.h` from FreeBSD

Hm.. Cool, but  I see it includes a "<sys/cdefs.h>" (which may or may not work when compiling from a windows host ?)
I notice ARRAY_SIZE in there, but what else do we need from that file ?

>     * use `__algined()` instead of `__ALIGN_BEGIN` and `__ALIGN_END`

+1

marckleinebudde:
> Hm.. Cool, but I see it includes a "<sys/cdefs.h>" (which may or may not work when compiling from a windows host ?) I notice ARRAY_SIZE in there, but what else do we need from that file ?

Removed `cdefs.h`, don't know if this would fail on Windows :)

The `barrier`, `READ_ONCE` and `WRITE_ONCE` is needed for the Linux lists, so I don't want to remove this and add back later.

+ rebased to latest master.

marckleinebudde:
Feel free to merge.

marckleinebudde:
Dammit, didn't want to closed it...

fenugrec:
thanks !

PR #133, opened 2022-11-13
editorconfig: add minimal editorconfig

marckleinebudde:
GH parses the `.editorconfig` for displaying code. Add minimal editorconfig to display our code, the STM code looks good at first glance, too.


marckleinebudde:
FTR: configure git's pager for tab with of 4:

```git config core.pager 'less -x1,5'```

configure text mode interface tig:
```git config tig.tab-size '4'```

PR #138, opened 2022-11-16
Fix the standby pin set/reset to use the active high/low define

ryedwards:
Code was ignoring the active flag as defined in config.h.  The flag was used in the GPIO init routines but not in can.c where it is used during can_enable and can_disable.

### DIFF
--- a/src/can.c
+++ b/src/can.c
@@ -150,7 +150,7 @@ void can_enable(can_data_t *hcan, bool loop_back, bool listen_only, bool one_sho
 	can->FMR &= ~CAN_FMR_FINIT;
 
 #ifdef nCANSTBY_Pin
-	HAL_GPIO_WritePin(nCANSTBY_Port, nCANSTBY_Pin, GPIO_PIN_SET);
+	HAL_GPIO_WritePin(nCANSTBY_Port, nCANSTBY_Pin, nCANSTBY_Active_High == 0 ? GPIO_PIN_RESET : GPIO_PIN_RESET);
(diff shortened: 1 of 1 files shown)


src/can.c
```diff
@@ -150,15 +150,15 @@ void can_enable(can_data_t *hcan, bool loop_back, bool listen_only, bool one_sho
 	can->FMR &= ~CAN_FMR_FINIT;
 
 #ifdef nCANSTBY_Pin
-	HAL_GPIO_WritePin(nCANSTBY_Port, nCANSTBY_Pin, GPIO_PIN_SET);
+	HAL_GPIO_WritePin(nCANSTBY_Port, nCANSTBY_Pin, nCANSTBY_Active_High == 0 ? GPIO_PIN_RESET : GPIO_PIN_RESET);
```
fenugrec, line 153: Umm, this would only ever do GPIO_PIN_RESET ?

Also, we have a macro `GPIO_INIT_STATE` in gpio.c which could be moved to gpio.h and used here FGJ

fenugrec (changes requested):

src/can.c
ryedwards: Agree...not sure what happened here.  Needs to be updated.

ryedwards: Fixed in new commit.

fenugrec:
Thanks. I rewrote it slightly differently though in f4f1586dbf563b4850d5b52e99aac06f0311403e

PR #139, opened 2022-11-17
add support for multichannel and CAN-FD

marckleinebudde:
This PR should add multichannel and CAN-FD support. Currently it's WIP.


src/usbd_gs_can.c
```diff
@@ -606,74 +437,286 @@ static uint8_t USBD_GS_CAN_Setup(USBD_HandleTypeDef *pdev, USBD_SetupReqTypedef
...
+							 led_mode_normal : led_mode_off);
+			}
+			break;
+
+		case GS_USB_BREQ_SET_TERMINATION:
+			if (get_term(req->wValue) != GS_CAN_TERMINATION_UNSUPPORTED) {
+				memcpy(&param_u32, hcan->ep0_buf, sizeof(param_u32));
+				if (set_term(req->wValue, param_u32) == GS_CAN_TERMINATION_UNSUPPORTED) {
+					USBD_CtlError(pdev, req);
+				}
```
ryedwards, line 524: Termination GPIO is channel based.  Should we stub that out here as part of the multichannel change?

src/usbd_gs_can.c
```diff
@@ -477,45 +306,59 @@ static uint8_t USBD_GS_CAN_DFU_Request(USBD_HandleTypeDef *pdev, USBD_SetupReqTy
...
+	 * channel number in req->wValue. So check req->wValue for valid
+	 * CAN channel.
+	 *
+	 */
+	if (!(req->bRequest == GS_USB_BREQ_HOST_FORMAT ||
+		  req->bRequest == GS_USB_BREQ_DEVICE_CONFIG)) {
+		channel = USBD_GS_CAN_get_channel(hcan, req->wValue);
+		if (!channel)
+			goto out_fail;
+	}
```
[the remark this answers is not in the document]
fenugrec: it's so you don't have to rewrite multiple times (i.e. error-prone) a common set of statements that clean up on return. I'm ok with this type of goto usage.

src/main.c
```diff
@@ -28,143 +28,138 @@ THE SOFTWARE.
...
-#ifdef CAN_S_GPIO_Port
-	HAL_GPIO_WritePin(CAN_S_GPIO_Port, CAN_S_Pin, GPIO_PIN_RESET);
-#endif
-
 	while (1) {
-		struct gs_host_frame *frame = queue_pop_front(q_from_host);
-		if (frame != 0) { // send can message from host
-			if (can_send(&hCAN, frame)) {
+		can_data_t *channel = &hGS_CAN.channels[0];
+		struct gs_host_frame_object *frame_object;
```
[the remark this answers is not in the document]
marckleinebudde: no - multichannel will come later

src/usbd_gs_can.c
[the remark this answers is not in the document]
ryedwards: I was just beat up side the head years ago by a senior developer.  Scars haven't healed.  I think he's retired now.

src/main.c
```diff
@@ -28,143 +28,138 @@ THE SOFTWARE.
...
-	q_frame_pool = queue_create(CAN_QUEUE_SIZE);
-	q_from_host  = queue_create(CAN_QUEUE_SIZE);
-	q_to_host    = queue_create(CAN_QUEUE_SIZE);
-	assert_basic(q_frame_pool && q_from_host && q_to_host);
+		led_set_mode(&channel->leds, led_mode_off);
 
-	struct gs_host_frame *msgbuf = calloc(CAN_QUEUE_SIZE, sizeof(struct gs_host_frame));
-	assert_basic(msgbuf);
+		can_init(channel, CAN_INTERFACE);
+		can_disable(channel);
```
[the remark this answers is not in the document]
marckleinebudde: You mean a read only array of structs, with a struct describing the HW configuration of a single channel. Sounds good. Which information do we need per channel?

src/usbd_gs_can.c
marckleinebudde: Should come after we thought about per channel configuration.

src/main.c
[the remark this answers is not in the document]
marckleinebudde: I've pushed an update: https://github.com/candle-usb/candleLight_fw/pull/139/commits/a390079bb10b6a2f2ccc1ed7b15d7e151ecc86d2

src/main.c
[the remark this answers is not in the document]
marckleinebudde: Here's an update: https://github.com/candle-usb/candleLight_fw/pull/139/commits/a390079bb10b6a2f2ccc1ed7b15d7e151ecc86d2

marckleinebudde:
It's not yet stable, `canfdtest -vg can0` stops after some minutes.

include/can-common.h
```diff
@@ -0,0 +1,36 @@
+/*
```
ryedwards, line 1: Should we follow the naming convention for the project and name the file "can_common.h" and "can_common.c"?

marckleinebudde: Will fix, thanks for pointing out!

marckleinebudde: done

marckleinebudde:
changes:
* renames `can-common.*` to `can_common.*`

marckleinebudde:
changes:
- fixed includes *cough*

marckleinebudde:
FYI: https://github.com/lichtfeind will now work on STM32G0 support.

include/usbd_gs_can.h
```diff
@@ -48,27 +54,35 @@ THE SOFTWARE.
 
 extern USBD_ClassTypeDef USBD_GS_CAN;
 
+#ifdef CONFIG_CANFD
+#define GS_HOST_FRAME_SIZE struct_size((struct gs_host_frame *)NULL, canfd_ts, 1)
+#else
+#define GS_HOST_FRAME_SIZE struct_size((struct gs_host_frame *)NULL, classic_can_ts, 1)
+#endif
+
+static_assert(CAN_DATA_MAX_PACKET_SIZE >= GS_HOST_FRAME_SIZE);
```
ryedwards, line 63: This line is causing the following compiler errors:
```
error: expected declaration specifiers or '...' before numeric constant
   43 | #define CAN_DATA_MAX_PACKET_SIZE 32    /* Endpoint IN & OUT Packet size */
```
and
```
note: in expansion of macro 'CAN_DATA_MAX_PACKET_SIZE'
   63 | static_assert(CAN_DATA_MAX_PACKET_SIZE >= GS_HOST_FRAME_SIZE);
```

marckleinebudde: fixed

fenugrec:
I understand this is still WIP, but if there's a subset that you think is ready to be merged (I'm thinking of e.g. the dfu changes and some of the earlier, more general changes) we could do that ; I think cherry-picking a few commits to master then rebasing this branch should be doable as a fairly clean , low-entropy operation

marckleinebudde:
@fenugrec https://github.com/candle-usb/candleLight_fw/pull/156 is ready to merge!

cym-bmeisels:
Hi,
Classic CAN works for me but I am having trouble using FD. Can you share your setup (Kernel version and command line input)?

cym-bmeisels:
@marckleinebudde Thanks for your response.
I have been using https://github.com/microsoft/WSL2-Linux-Kernel/tree/linux-msft-wsl-6.1.21.2 . Would you happen to know if it is too old?

marckleinebudde:
According to https://github.com/microsoft/WSL2-Linux-Kernel/commits/linux-msft-wsl-6.1.21.2/drivers/net/can/usb/gs_usb.c already supports CAN-FD. As said in my previous comment, make sure to use the latest firmware, there was a bug in some older version.

cym-bmeisels:
I must have misread. I thought you said the issue was in the kernel. Will update the firmware and try again.
Thanks!

marckleinebudde:
Actually I asked for both :)

src/can/m_can.c
```diff
@@ -0,0 +1,383 @@
...
+
+// bit timing constraints
+const struct gs_device_bt_const CAN_btconst = {
+	.feature =
+		GS_CAN_FEATURE_LISTEN_ONLY |
+		GS_CAN_FEATURE_LOOP_BACK |
+		GS_CAN_FEATURE_HW_TIMESTAMP |
+		GS_CAN_FEATURE_IDENTIFY |
+		GS_CAN_FEATURE_PAD_PKTS_TO_MAX_PKT_SIZE |
+		(IS_ENABLED(CONFIG_CANFD) ?
```
bmeisels, line 41: This doesn't seem to work

bmeisels:
I finally have a normal debug setup and have narrowed it down to m_can.,c:41 not working for my custom target.
The rest of the config seems to work and when I remove the IS_ENABLED macro all seems to work perfectly.

This is my board config:

```` C
#elif defined(BOARD_cymocan)
	#define USBD_PRODUCT_STRING_FS	 (uint8_t *)"CYMOCAN gs_usb"
	#define USBD_MANUFACTURER_STRING (uint8_t *)"CYMOTIVE"
	#define DFU_INTERFACE_STRING_FS	 (uint8_t *)"CYMOCAN firmware upgrade interface"

	#define TIM2_CLOCK_SPEED		 64000000

	#define CAN_CLOCK_SPEED			 40000000
	#define NUM_CAN_CHANNEL			 2
	#define CONFIG_CANFD			 1
````

**Comment by marckleinebudde** [comment]
I've imported the zehpyr version of IS_ENABLED, and that seems to work. I've force pushed the branch, please test.

**Comment by bmeisels** [comment]
Thanks,
This fixes the issue.

Is there anything I can do to help this get merged?
Are there any blockers?

**Comment by fenugrec** [comment]
> Are there any blockers?

Mainly one : it's still described as "Currently it's WIP." in the first post !

**Comment by marckleinebudde** [comment]
I'm still not satisfied with the LED handling.

**Comment by bmeisels** [comment]
Is this anything I could help with?

**Comment by sammcj** [comment]
Keen to see this merged in!

**Comment by ryedwards** [comment]
> I'm still not satisfied with the LED handling.

You are always welcome to use my LED implementation - follows a similar architecture where each LED has a handle and is managed by a single period function call.

https://github.com/ryedwards/budgetcan_fw/blob/main/Core/Src/led.c

**Comment by fenugrec** [comment]
@marckleinebudde I'm thinking of merging part of this (early commits up to 6af7be218a8085072470f5fb98ceee0b700b0914 `treewide: sort includes`  as they look fairly harmless. Will make subsequent reviews less intimidating than a stack of 83 commits. Unless you plan on refactoring those early commits eventually ?

**Comment by marckleinebudde** [comment]
You can merge up to 99973e94e4b1 ("usbd_gs_can: add support for GS_USB_BREQ_DATA_BITTIMING and GS_USB_BREQ_BT_CONST_EXT"), or smaller steps if you don't want to review that much.

**Comment by fenugrec** [comment]
In 91f61ffc7364b8bfc367f7479e05989a8a4f33b2 `compiler.h: add static_assert() ` , I'm not sure I like how it redefines `static_assert` to use varargs; C23 defines static_assert differently (C11 didn't) and I expect we may run into problems eventually. Any reason not to use something like STATIC_ASSERT() where it's immediately obvious we're using a macro that will not collide with any compiler, present or future ?

**Comment by marckleinebudde** [comment]
The code doesn't use `static_assert()` any more. I'll remove that patch.

Thanks for the review.

**Comment by marckleinebudde** [comment]
I've removed the `static_assert()` patch.

**Comment by marckleinebudde** [comment]
> the next one re BUILD_BUG and CTASSERT :
> 
> +`#define BUILD_BUG()						do { CTASSERT(0); } while (0)`
> 
> don't seem to be used either ? I'll defer them until they're actually used. Also, currently I don't think anything provides CTASSERT so it would probably not work ?

Granted, it's not used. The point of `BUILD_BUG` and `BUILD_BUG_ON()` that the build fails. I'll remove them for now.

> I may also defer these and the "add min/max" macros until they're needed. Should be easier to cherry-pick / reorder later

The `max5()` thingy is actually used. I'll reorder the patch stack.

**Comment by fenugrec** [comment]
> The `max5()` thingy is actually used. I'll reorder the patch stack.

hm, is it perhaps in some code that then gets deleted ? On the tip of this PR,
```
$ grep "min5" * -R
include/compiler.h:#define min5(a, b, c, d, e) min3(min(a, b), min(c, d), e)
```

and `git log -S` only returns the one commit where they are added.

In any case, thanks for your continued patience and work on this branch. We'll get there...

**Comment by marckleinebudde** [comment]
removed `BUILD_BUG` and moved `max5()` prior to the patch that's using it

**Comment by fenugrec** [comment]
Merged a few more commits.
Next question : in f2a071ac48f1c432706461da41342f09b9cbf70c , you are reordering switch statements, but it's not clear what order you're aiming for : it's not alphabetical, and doesn't seem to follow `enum gs_usb_breq` either ?

**Comment by marckleinebudde** [comment]
In USBD_GS_CAN_Config_Request(), sort the switch statement by GS_USB_BREQ vlaue as much as possible, but don't duplicate code.
    
First place the "Host -> Device" calls which use USBD_CtlPrepareRx(). Followed by the "Device -> Host" calls which are using USBD_CtlSendData().

I've updated the commit message accordingly.

Hope this makes more sense now. 😸

**Comment by pgreenland** [comment]
@marckleinebudde If it helps....

I had a STM32G0 dual channel CAN adapter in need of software, see my recent blog post for more details: https://www.quantulum.co.uk/blog/adventures-in-can-fd/

Forked your repo and implemented the board file for my board (https://github.com/pgreenland/candleLight_fw/tree/usbtocanfd_v1).

It's working great. Been running canfdload between a pair of adapters running the same build. 2M+ messages transferred with no errors.

Have tested using both CAN 2.0 and CAN FD with bit rate switching at 1Mbps nominal / 8Mbps data.

LEDs and identify requests for each channel are looking sensible.

Per channel termination control is working fine.

The only tweaks I made, enabling automatic transceiver delay compensation. Without this I wasn't able to run at data bitrates above 2Mbps, see: https://github.com/pgreenland/candleLight_fw/commit/76d0a3bfc4dea308328aeda64b75f6d72a2408e2

Tacked in an option to support an external clock, rather than the onboard RC oscillator. When designing my board I'd read that the onboard oscillator wouldn't be stable enough for CAN FD at high speeds. It seemed to be working ok, but I only tried a loopback test between controllers on the same adapter, hiding any clock issues. All testing between adapters has been performed using the external clock config here: https://github.com/pgreenland/candleLight_fw/commit/6c24607b2b04c873ba6e57c1d46492ef9e312423

As a random aside, I was surprised to find the device side gs_can driver echos back the whole transmitted can frame, when all the host really needs is the echo_id to pair up with its original request. For CAN FD this becomes even more of a problem as the 80 byte frames take up more USB bandwidth than the smaller CAN 2.0 ones, requiring two transfers. Excluding the payload from echo replies, increased my CAN FD transmission rate from 5000 to 6000 msg/s. gs_can doesn't even notice or care :-).

Thanks,

Phil

**Comment by fenugrec** [comment]
c85a873e066ff40242e27579c5c591e410131039 casting`hcan->ep0_buf` directly as a struct... not convinced of the advantage of doing that. We are already doing similar casts with other structs so "it should be fine" I guess. This will break spectacularly if `ep0_buf` is ever unaligned. We should probably add an align(4) directive on that or at the very least a comment to make sure it stays at the beginning of the struct.

**Comment by marckleinebudde** [comment]
Added a `__aligned(4)` to `USBD_GS_CAN_HandleTypeDef::ep0_buf`.

**Comment by fenugrec** [comment]
I will need to look closer at 70962b7ec0dec8ffaa8c2c25246bd085b8e32027 "usbd_gs_can: clean up send to host path"  since it introduces some non-trivial changes. What was the rationale for that ?

**Comment by marckleinebudde** [comment]
Luckily I added a patch description:

    The USBD_GS_CAN_SendToHost() function is used to send to struct
    gs_host_frame_object to the host.
    
    Until this patch, after the sending process has been started, the
    outgoing frame object is already added to the list of free objects and
    the variable USBD_GS_CAN_HandleTypeDef::TxState is used to track if
    the transfer to the host is in progress.

What I didn't like about this approach is, that the frame object as soon as transmission is _started_ is added to the list of free objects. But we need to track if a transmission is ongoing. Therefor the variable USBD_GS_CAN_HandleTypeDef::TxState is used.
    
    Instead, hold the outgoing object in
    USBD_GS_CAN_HandleTypeDef::to_host_buf and move it to the free list
    after the transfer is finished in USBD_GS_CAN_DataIn(). Use this to
    track if a transfer is ongoing.

The patch replaces the USBD_GS_CAN_HandleTypeDef::TxState by USBD_GS_CAN_HandleTypeDef::to_host_buf, which points to the frame object which is currently transferred. Also wait until the transfer is done, then move the frame object from USBD_GS_CAN_HandleTypeDef::to_host_buf to the list of free objects.

**Comment by fenugrec** [comment]
Finally got some hardware wired up again.

On master 3e4883c74f0183b50815d524b92b259897a8b14d , if I pick my best combination of USB ports, at 1Mbps I  get around 19000 frames/s with `cangen can0 -g 0 -p 10 -I 0x555` (specifying this identifier should reduce bit-stuffing overhead), that is 80% bus load  (from `canbusload can0@1000000 -r -e`  )

With just  c23453b531a1198e3de5db2d000f9ab1ac6ca7cf  (removing the `TxState` flag ), this drops to 18650 frmes/s . Testing a few commits later ff9c6aa99ddb5bf8266d12b71b2b086c9e1d1552 drops to 16700. Not a huge penalty if we're gaining something in exchange, but I'm not sure what exactly we're gaining.


> What I didn't like about this approach is, that the frame object as soon as transmission is _started_ is added to the list of free objects.

At the risk of looking dense, I would like clarification on why this was a problem ? My understanding is that from the moment we call USBD_LL_Transmit, the frame is copied to the USB PMA buffer and therefore we don't need to keep the source data around.
Are we concerned with 
- any problems that would cause that frame to become lost and never transmitted to host ?
- eventual use of double-buffering (actually it seems to be partially enabled already, didn't dig this) and/or multiple-packet transfers where it would be important to *not* put the frame back into the pool until the transfer is done ?
- some other scenario that I'm not seeing, possibly related to multichannel ?

(For reference, in case the commit hashes change , I was testing this series of commits :
```
    usbd_gs_can: USBD_GS_CAN_SendFrame(): call USBD_GS_CAN_Transmit() with IRQs enabled
    usbd_gs_can: USBD_GS_CAN_ReceiveFromHost(): call USBD_GS_CAN_PrepareReceive() with IRQs enabled
    usbd_gs_can: USBD_GS_CAN_ReceiveFromHost(): remove some levels of indention
    usbd_gs_can: clean up send to host path
```

**Comment by akohlsmith** [comment]
Is it possible to build this for canable2? I don't see any STM32G4 support and `make help` doesn't show me anything obvious.

**Comment by marckleinebudde** [comment]
Hey @bmeisels,

the next steps would be to address @fenugrec review in https://github.com/candle-usb/candleLight_fw/pull/139#issuecomment-1817686551, i.e. debug the performance regression.

**Comment by ryedwards** [comment]
So, something interesting I found on another project I am working on that uses multichannel CAN that will need to be addressed.  We had users complain that when they enabled the 2nd channel on the board CAN on the 1st channel would stop working.  We found that users were enabling channel 2 with nothing attached so the "from host" buffer would fill with channel 2 frames.  Of course they wanted this treated as a "bug" vs. user error.  Not sure if it's something we need to consider in this design (different queues for each channel?)

**Comment by akohlsmith** [comment]
that definitely sounds like a bug @ryedwards; I would expect the two CAN channels to have separate queues and that a failure or misconfiguration on one channel would not affect the operation of the other.

**Comment by ryedwards** [comment]
I just realized I posted this in the dead multichannel PR - I'll move to the current one.

**Comment by fenugrec** [comment]
I've decided to push the commits reworking the txstate flag logic, despite the performance hit on F0. I think that change was going to be needed anyway if we rework the echo frame logic; hopefully we'll regain some performance then

**Comment by marckleinebudde** [comment]
Hello @fenugrec,

thanks for pulling in the changes. What are the next steps?

regards,
Marc

**Comment by fenugrec** [comment]
> thanks for pulling in the changes. What are the next steps?

One of the commits adds `int can_check_bittiming(` and I keep forgetting to ask if there is a future reason why this returns int instead of `bool`  since it's only used for validation ?

Next, I was comparing the branches here and in 176 and it looks like they differ at some point in the m_can commits so I was going to ask which branch to proceed with.

File: src/boards/g0b1-nucleo_g0b1re.c [hunk 8]
```diff
@@ -0,0 +1,141 @@
...
+
+	HAL_GPIO_WritePin(LEDRX_GPIO_Port, LEDRX_Pin, GPIO_INIT_STATE(LEDRX_Active_High));
+	GPIO_InitStruct.Pin = LEDRX_Pin;
+	GPIO_InitStruct.Mode = LEDRX_Mode;
+	GPIO_InitStruct.Pull = GPIO_NOPULL;
+	GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_LOW;
+	HAL_GPIO_Init(LEDRX_GPIO_Port, &GPIO_InitStruct);
+
+	/* FDCAN */
+
```
IvoBCD, line 63: FWIW - To get this to work with Adafruit "CAN Pal" TJA1051T/3 breakout board( https://www.adafruit.com/product/5708 ) , I needed to connected the transceiver silent (SLNT) pin to GND.

marckleinebudde: In your board file you can implement the `struct BoardConfig::phy_power_set` callback and pull the transceiver's silent pin into the correct direction depending on the value of the `enable` parameter.

trentatsa:
@marckleinebudde, is CANFD fully supported on G0 based devices ? I have recently purchased candleLight FD from linux automation and im looking for something that works out the box.

marckleinebudde:
> @marckleinebudde, is CANFD fully supported on G0 based devices ? I have recently purchased candleLight FD from linux automation and im looking for something that works out the box.

The device comes pre-flashed with a working firmware.

pramilo:
Hi all

Can you pls clarify if this has finally been merged to main_dev - or - if we need to get code from [marckleinebudde]'s (https://github.com/marckleinebudde) fork?

The discussion seems to indicate support for the G0 is mature, but it's unclear if/what's been merged to this main branch.

Thanks,

marckleinebudde:
It's not mainline. Please use https://github.com/marckleinebudde/candleLight_fw/tree/multichannel for now

indrapippo:
> It's not mainline. Please use https://github.com/marckleinebudde/candleLight_fw/tree/multichannel for now

Hi there! Multichannel means that this will rise can0 and can1 interfaces?

marckleinebudde:
This means on STM32G0B1 devices you have 2 interfaces, `can0` and `can1`. You need 2 PHYs on your HW for them to work.

PR #141, opened 2022-11-17
startup: get rid of zero table and fix copy_table

marckleinebudde:
Works for me on the STM32F072, can someone test the 042?

Should fix the this issue: https://github.com/candle-usb/candleLight_fw/pull/135#issuecomment-1317912852

### DIFF
--- a/ldscripts/STM32F042X6.ld
+++ b/ldscripts/STM32F042X6.ld
@@ -62,13 +62,6 @@ SECTIONS
     __copy_table_end__ = .;
   } > FLASH
 
-  .zero.table :
-  {
-    . = ALIGN(4);
-    __zero_table_start__ = .;
-    __zero_table_end__ = .;
-  } > FLASH
-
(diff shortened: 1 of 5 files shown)


fenugrec:
haven't tested, but we risk having problems again because of how `wlen` is generated in the ldscripts :

```
    LONG (__etext)
    LONG (__data_start__)
    LONG (__data_end__ - __data_start__)
```

i.e. the above becomes the `copy_table_t` struct; needs a divide-by-4 there (I think those 3 symbols are guaranteed to be aligned-4  - didn't check yet

fenugrec:
Thanks.
Rebased, tested (briefly) on F042, and merged in 59896b65fdd588fed40cc4d9efec00d21f04ad31

sidenote - I will "soon" tweak the uncrustify.cfg to enforce { } around everything : if/else,  single-line for loops, etc.. It's a nasty trap.

PR #143, opened 2022-11-22
cmake: switch from -Os to -O2 optimization

marckleinebudde:
This increases the code size, but even on a STMF042 fits into the flash:

-Os
```
Memory region         Used Size  Region Size  %age Used
           FLASH:       14180 B        32 KB     43.27%
             RAM:        3744 B         6 KB     60.94%
```

-O2
```
Memory region         Used Size  Region Size  %age Used
           FLASH:       17284 B        32 KB     52.75%
             RAM:        3744 B         6 KB     60.94%
```

This optimization increases the max TX CAN bus load on a STM32F072 (1 MBit/s, DLC=1) from 77% to 84%.

For completeness:

-O3
```
Memory region         Used Size  Region Size  %age Used
           FLASH:       22568 B        32 KB     68.87%
             RAM:        3744 B         6 KB     60.94%
```
The max TX CAN bus load is 88%.

PR #144, opened 2022-11-23
LED performance rework

fenugrec:
Sidequest from #143 - we're spending way too much time updating LEDs.

The most gain here is obtained from calling led_update() less often (every 10ms with this patch).

The commit changing the implementation of `led_indicate_trx()` showed surprisingly marginal gains - about 0.5% with a `cangen -L0 -g0 -p10 can0` benchmark.

can busload goes from ~ 12900 to 13120 frames/s with this PR.

### DIFF
--- a/src/led.c
+++ b/src/led.c
@@ -85,23 +85,27 @@ void led_run_sequence(led_data_t *leds, const led_seq_step_t *sequence, int32_t 
 	led_update(leds);
 }
 
-void led_indicate_trx(led_data_t *leds, led_num_t num)
-{
-	uint32_t now = HAL_GetTick();
-	led_state_t *led = &leds->led_state[num];
+void led_indicate_trx(led_data_t *leds, led_num_t num) {
+	leds->led_state[num].blink_request = 1;
+}
 
-	if ( SEQ_ISPASSED(now, led->on_until) &&
-		 SEQ_ISPASSED(now, led->off_until) ) {
-		led->off_until = now + 30;
-		led->on_until = now + 45;
+static void led_trx_blinker(led_state_t *ledstate, uint32_t now) {
+	if ( SEQ_ISPASSED(now, ledstate->on_until) &&
+		 SEQ_ISPASSED(now, ledstate->off_until) ) {
+		ledstate->off_until = now + 30;
+		ledstate->on_until = now + 45;
 	}
-
-	led_update(leds);
 }
 
 static void led_update_normal_mode(led_state_t *led)
 {
 	uint32_t now = HAL_GetTick();
+
+	if (led->blink_request) {
+		led->blink_request = 0;
+		led_trx_blinker(led, now);
+	}
+
 	led_set(led, SEQ_ISPASSED(now, led->off_until));
 }
 
--- a/src/led.c
+++ b/src/led.c
@@ -137,6 +137,13 @@ static void led_update_sequence(led_data_t *leds)
 
 void led_update(led_data_t *leds)
 {
+	static uint32_t next_update = 0;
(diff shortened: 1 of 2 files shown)


src/led.c
```diff
@@ -75,31 +77,37 @@ static uint32_t led_set_sequence_step(led_data_t *leds, uint32_t step_num)
...
+		ledstate->off_until = now + 30;
+		ledstate->on_until = now + 45;
+	}
 }
 
 static void led_update_normal_mode(led_state_t *led)
 {
 	uint32_t now = HAL_GetTick();
+
+	led_trx_blinker(led, now);
```
marckleinebudde, line 109: nitpick: I would move the check for `led->blink_request` from the callee to the caller:

```suggestion
	if (led->blink_request) {
		led_trx_blinker(led, now);
	}
```

src/led.c
```diff
@@ -75,31 +77,37 @@ static uint32_t led_set_sequence_step(led_data_t *leds, uint32_t step_num)
...
+	if ( SEQ_ISPASSED(now, ledstate->on_until) &&
+		 SEQ_ISPASSED(now, ledstate->off_until) ) {
+		ledstate->off_until = now + 30;
+		ledstate->on_until = now + 45;
+	}
 }
 
 static void led_update_normal_mode(led_state_t *led)
 {
 	uint32_t now = HAL_GetTick();
```
marckleinebudde, line 107: nitpick: you can pass `now` from the caller:

```suggestion
static void led_update_normal_mode(led_state_t *led, uint32_t now)
{
```

src/led.c
```diff
@@ -75,31 +77,37 @@ static uint32_t led_set_sequence_step(led_data_t *leds, uint32_t step_num)
...
+}
 
-	if ( SEQ_ISPASSED(now, led->on_until) &&
-		 SEQ_ISPASSED(now, led->off_until) ) {
-		led->off_until = now + 30;
-		led->on_until = now + 45;
+static void led_trx_blinker(led_state_t *ledstate, uint32_t now) {
+	if (!ledstate->blink_request) {
+		return;
 	}
```
marckleinebudde, line 94: nitpick: I would move the check for led->blink_request from the callee to the caller:

```suggestion
```

src/led.c
marckleinebudde: Seem I cannot provide a suggestion where `led_update_normal_mode()` is called to add the `now` parameter 😿

marckleinebudde (commented):
Looks good! Minor nitpicks only.

src/led.c
fenugrec: That's something I sometimes hesitate about - hiding a simple run/return condition like this inside the func, or have it at the caller level. Doing it from the caller can become heavy if there are multiple call sites (or if the caller shouldn't know about the details of the conditions), but on the other hand it's more obvious that calling the function will do something and shouldn't always be called.

Here it,s all static funcs so both approaches would probably compile to the same code, so I don't feel strongly either way. What is your reasoning for this type of use case ?

src/led.c
fenugrec: Agreed

src/led.c
marckleinebudde: IMHO it's clearer to read, the function is called `led_update_normal_mode()` so blinking is a special case of normal mode. We only call `blinker` if some condition (`led->blink_request`) is set.

If you are not coming from the blinker use case, but "turn off the LED", then I'm wandering why is `blinker` called unconditionally.

fenugrec: Fair. Will change.

marckleinebudde:
looks good to me!

PR #147, opened 2022-11-28
override __register_exitproc()

fenugrec:
atexit() is meaningless on this kind of embedded fw. Overriding this symbol saves a bunch of memory (384 B flash on cannette).

Ref:
https://sourceware.org/git/?p=newlib-cygwin.git;a=blob_plain;f=newlib/libc/stdlib/__atexit.c;hb=HEAD

### DIFF
--- a/src/startup.c
+++ b/src/startup.c
@@ -26,6 +26,9 @@ void Reset_Handler(void)
 	_start();
 }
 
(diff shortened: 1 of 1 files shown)


marckleinebudde:
Looks good to me, tested on candlelight.

PR #148, opened 2022-12-05
RFC: Split device-specific code to different files

fenugrec:
Draft, still thinking about this. Comments welcome.

Since we don't have tons of device-specific code yet, I was putting all the code into a single file per mcu type. This may become impractical if

- mcus from a same family have different CAN/USB peripherals , or more fundamental differences
- adding tons of unrelated functions like what I've started here, with CAN and system clock config in the same file, may become messy. But maybe still manageable : only bits left are GPIO and USB.

GPIO may end up in a board-specific hierarchy instead, still thinking about this.


include/device.h
```diff
@@ -0,0 +1,6 @@
+#pragma once
```
marckleinebudde, line 1: Nitpick: I would add the common copyright header here.

marckleinebudde:
Haven't tested but looks good!

PR #149, opened 2022-12-10
add xCANFD board config info

ConvertDevice:


marckleinebudde:
Please rebase your code to current master, so that it doesn't contain any merges. The last patch not only adds support for the CONVERTDEVICE_xCANFD, but also renames the CONVERTDEVICE_xCAN01 to CONVERTDEVICE_xCAN. Please make this 2 separate patches.

FYI: Support for the STM32g0b not not merged to mainline, yet. Several people are working on this.

PR #152, opened 2022-12-15
WIP: Add support for STM32G0 with can split in separate file 

lichtfeind:
This is based on @ryedwards  pull request #126  but tries to separate out the CAN relater changes into a separate file.

I am already on holiday so @marckleinebudde will continue to work in this.

This is developed on an NUCLEO-G0B1RE.
And at the moment does not work yet.
TX of CAN frames does not work and it only receives once.

TODO:
  - Fix CAN
  - rebase on multichannel #139 
  - Add some board definitions to support different pinmux configs between boards.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,7 +65,7 @@ set(
 	include/usbd_gs_can.h src/usbd_gs_can.c
 	src/usbd_conf.c
 
-	include/can.h src/can.c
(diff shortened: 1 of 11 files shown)


marckleinebudde:
I've included the stm32g0b1 support in https://github.com/marckleinebudde/candleLight_fw/tree/multichannel, please close.

PR #155, opened 2023-01-16
bxcan: can_enable(): fix disabling of standby pin

marckleinebudde:
The binary inverse of 0x1 is 0xfe not 0x0. Fix this by using a logic negation instead.

Fixes: f4f1586dbf56 ("can.c: use GPIO_INIT_STATE to set CANSTBY pin")


fenugrec:
Thanks ! dumb mistake...

PR #156, opened 2023-01-18
Fix stm32g0b USB stall

marckleinebudde:
Under high CAN-TX load the endpoint on the stm32g0b will stall, this PR fixes the issue.

Issue #82, opened 2022-01-03
RFC: consider pulling in ST HAL + CMSIS as submodules instead of local copy


fenugrec:
drawback : those repos, even when shallow cloned, are huge. [CubeF0](https://github.com/STMicroelectronics/STM32CubeF0/) is > 300MB,[ HAL alone](https://github.com/STMicroelectronics/stm32f0xx_hal_driver) is almost 200 MB.

Daniel-Trevitz:
Tiny USB is using: 
https://github.com/STMicroelectronics/stm32f0xx_hal_driver.git
https://github.com/STMicroelectronics/stm32f4xx_hal_driver.git
https://github.com/STMicroelectronics/cmsis_device_f0.git
https://github.com/STMicroelectronics/cmsis_device_f4.git

When checked out they are currently 40MiB for the cmsis set, and 20MiB for the stm set.
--depth=1 clone only drops 2MB for a total of 58MiB

Our paired-down version of all 4 combined is 6.8MiB, and includes our cmake support.

fenugrec:
ah, this could help too :

>$  man gitmodules
> ...
>  submodule.<name>.shallow
>            When set to true, a clone of this submodule will be performed as a shallow clone (with a history depth of
>            1) unless the user explicitly asks for a non-shallow clone.
>

PR #157, opened 2023-01-18
STM32 HAL: update and use as submodules instead

fenugrec:
Will make it easier to track upstream bugfixes. Submodules are configured with "--depth 1" to make retrieval fast.

Partial fix for GH #82.

Version bumps :
F0xx : 1.7.5 -> 1.7.6
F4xx : 1.7.13 -> 1.8.1
G0xx : 1.4.5 (no change)

Problems :

- [ ] situations like #156 where we want to patch HAL librairies locally : not sure how to handle that.

- [ ] CI workflow needs to be adjusted


marckleinebudde:
With https://github.com/candle-usb/candleLight_fw/pull/227 the HALs are updated to their latest commits. I've adopted to structure of the upstream HALs, so future updates should be easier. Can we close this issue?

PR #160, opened 2023-04-22
gitignore: add .idea/ to ignored files

gemesa:
I am using CLion IDE which generates `.idea` for each project, similarly to VS Code `.vscode`. Can we add it to the ignore list?


fenugrec:
rebased & merged in 49eb5da92ac937c40ba33869793c4149802deee5

Issue #106, opened 2022-08-28
recent gcc complains about "LOAD segment with RWX permissions"


fenugrec:
`/usr/lib/gcc/arm-none-eabi/12.2.0/../../../../arm-none-eabi/bin/ld: warning: cantact_fw has a LOAD segment with RWX permissions`

Trying the naive fix of editing the .ldscripts RAM section as

`RAM   (rw) : ORIGIN = 0x20000000, LENGTH = 6K`

has no effect. 

A search quickly indicates we're not the only ones seeing this :

https://stackoverflow.com/questions/73429929/gnu-linker-elf-has-a-load-segment-with-rwx-permissions-embedded-arm-project

I haven't had time to dig any further.
I'm also fairly certain this can be disregarded on baremetal, non-MMU firmware like this.

marckleinebudde:
FYI: https://lore.kernel.org/all/20220810222442.2296651-1-ndesaulniers@google.com

fenugrec:
can we add `--no-warn-rwx-segments`systematically or would that break on older binutils versions ?

fenugrec:
Maybe cmake has a mechanism to check for ld-options ? 
Projects using plain makefiles have to resort to this kind of script

https://github.com/OP-TEE/optee_os/blob/203ee23d005b2cec2f21b5de334c5a246be32599/mk/cc-option.mk

marckleinebudde:
https://gist.github.com/multiplemonomials/9cc2c4343a4bdd93689520ce64098b84

PR #167, opened 2023-05-11
cmake: silence "LOAD segment with RWX permissions" warning

marckleinebudde:
Link: https://github.com/candle-usb/candleLight_fw/issues/106


marckleinebudde:
@fenugrec my compiler doesn't support `--no-warn-rwx-segments` and `cmake` doesn't use the new option. Does it fix the warning with your compiler?

marckleinebudde:
Debian stable ("bullseye") uses cmake-3.18, Debian oldstable ("buster") is still on cmake-3.13, but "buster-backports" has cmake-3.18, too.

fenugrec:
I think bumping required to 3.18 is reasonable; distros with really old cmake have a simple workaround of just reverting this commit.

marckleinebudde:
Thanks for testing! Fixed.

fenugrec:
> Fixed.

Thanks, youǜve fixed the passing-wrong-arg problem, but since the test compilation fails, cmake still thinks the flag is "not supported"

marckleinebudde:
Just noticed, too. But why does it say `cached: true`.

fenugrec:
> But why does it say `cached: true`.

AFAIK because https://cmake.org/cmake/help/latest/module/CheckLinkerFlag.html

>  Stores the result in an internal cache entry named <var>.

Issue #164, opened 2023-04-28
build warning: "_close is not implemented and will always fail"


fenugrec:
Nice, new warnings with arm-gcc 13.1.0 :

```
/usr/lib/gcc/arm-none-eabi/13.1.0/../../../../arm-none-eabi/bin/ld: /usr/lib/gcc/arm-none-eabi/13.1.0/../../../../arm-none-eabi/lib/thumb/v7e-m/nofp/libg_nano.a(libc_a-closer.o): in function `_close_r':
/build/arm-none-eabi-newlib/src/build-nano/arm-none-eabi/thumb/v7e-m/nofp/newlib/../../../../../../newlib-4.3.0.20230120/newlib/libc/reent/closer.c:47: warning: _close is not implemented and will always fail
/usr/lib/gcc/arm-none-eabi/13.1.0/../../../../arm-none-eabi/bin/ld: /usr/lib/gcc/arm-none-eabi/13.1.0/../../../../arm-none-eabi/lib/thumb/v7e-m/nofp/libg_nano.a(libc_a-lseekr.o): in function `_lseek_r':
/build/arm-none-eabi-newlib/src/build-nano/arm-none-eabi/thumb/v7e-m/nofp/newlib/../../../../../../newlib-4.3.0.20230120/newlib/libc/reent/lseekr.c:49: warning: _lseek is not implemented and will always fail
/usr/lib/gcc/arm-none-eabi/13.1.0/../../../../arm-none-eabi/bin/ld: /usr/lib/gcc/arm-none-eabi/13.1.0/../../../../arm-none-eabi/lib/thumb/v7e-m/nofp/libg_nano.a(libc_a-readr.o): in function `_read_r':
/build/arm-none-eabi-newlib/src/build-nano/arm-none-eabi/thumb/v7e-m/nofp/newlib/../../../../../../newlib-4.3.0.20230120/newlib/libc/reent/readr.c:49: warning: _read is not implemented and will always fail
/usr/lib/gcc/arm-none-eabi/13.1.0/../../../../arm-none-eabi/bin/ld: /usr/lib/gcc/arm-none-eabi/13.1.0/../../../../arm-none-eabi/lib/thumb/v7e-m/nofp/libg_nano.a(libc_a-writer.o): in function `_write_r':
[...]

PR #182, opened 2024-06-04
main: provide weak symbols for _close, _lseek, _read, _write

marckleinebudde:
This increases the size of the binary in a stm32f072 a bit:

before:
```
Memory region         Used Size  Region Size  %age Used
           FLASH:       17296 B       128 KB     13.20%
             RAM:        4268 B        16 KB     26.05%
```

after:
```
Memory region         Used Size  Region Size  %age Used
           FLASH:       17312 B       128 KB     13.21%
             RAM:        4268 B        16 KB     26.05%
```

Closes: https://github.com/candle-usb/candleLight_fw/issues/164

### DIFF
--- a/include/compiler.h
+++ b/include/compiler.h
@@ -53,14 +53,14 @@
 #define ACCESS_ONCE(x) (*(volatile __typeof(x) *)&(x))
 
 #define WRITE_ONCE(x,v) \
-	do { \
-		barrier(); \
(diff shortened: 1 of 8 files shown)


marckleinebudde:
Please merge https://github.com/candle-usb/candleLight_fw/pull/183 first, so that the uncrustify check is happy.

PR #183, opened 2024-06-04
uncrustify: update config file and run uncrustify

marckleinebudde:
Upgrade uncrustify's config file to the latest version.
Then uncrustify the header files in `include`.
Also I've found some unused defines in the `config.h`, remove them.
Remove comments about not connected silent pin and define `CONFIG_CANFD` to `1`, this way uncrustify produces a nice looking `config.h`.

### DIFF
--- a/include/compiler.h
+++ b/include/compiler.h
@@ -53,14 +53,14 @@
 #define ACCESS_ONCE(x) (*(volatile __typeof(x) *)&(x))
 
 #define WRITE_ONCE(x,v) \
-	do { \
-		barrier(); \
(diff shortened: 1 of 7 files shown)


fenugrec:
awesome. Let me know when it's ready (will probably do this, then 182, then 139, unless you have a different preferred order ?)

marckleinebudde:
This PR is ready, please pull it first, followed by https://github.com/candle-usb/candleLight_fw/pull/182. I'm still working on the multichannel stuff.

fenugrec:
Merged, thanks ! tested with uncrustify 0.77

PR #184, opened 2024-06-05
bxcan: advertise supported one-shot feature

de-vri-es:
This PR advertises one-shot mode as supported feature.

Without doing this, the gs_usb driver will report "operation not supported" when attempting to enable one-shot mode.

Tested locally with a candleLight, and it worked as expected :D


marckleinebudde:
Even if you enable one shot handling, you'll still get a TX complete under Linux. Keep that limitation in mind.

Looks good to me.

de-vri-es:
Yeah, there's no way to recognize transmission failure, but in my application this is still a lot better than potentially keeping the message around for hours until the bus connection is restored :)

fenugrec:
Rebased and merged, thanks.

> Keep that limitation in mind.

That reminds me we should document that, until the day we implement "echo frames after the message is sent"

PR #191, opened 2025-04-08
main: blink led wakeup pattern after initializing the USB Stack

### LINKED ISSUE #190: startup LED blinking might take too ling and lead to timeout of USB init
Cc: https://gist.github.com/dickenhobelix

dickenhobelix:
Blinking the LEDs for each channel can add up to quite a significant delay in the initialization, especially when initializing multiple CAN channels, but even for single-channel devices sometimes kernel errors are found on the host. For devices that cannot re-trigger USB enumeration due to hardwired pullups on the USB D+ line this may lead to timeouts and errors (Device not responding to setup address).

Blinking the LEDs after setting up the USB Peripheral and USB stack fixes this issue.

 - [x] buildable
 - [x] Tested on STM32F4VE Dev Board
 - [x] ran uncrustify

Fixes #190


fenugrec:
Also -  I agree with one of your comments in #190 , I see no reason to repeat the wake-up pattern for every channel : I just realized it's not blinking the _per-channel_ LEDs during init, but the same set of LEDs over and over. Feel free to amend this or add as an additional commit.

I am currently not in a position to test this, but when someone can test on F0 this should be good to merge.

fenugrec:
what I mean is, simply move
```
		/* nice wake-up pattern */
		for (uint8_t j = 0; j < 10; j++) {
			HAL_GPIO_TogglePin(led_config[LED_RX].port, led_config[LED_RX].pin);
			HAL_Delay(50);
			HAL_GPIO_TogglePin(led_config[LED_TX].port, led_config[LED_TX].pin);
		}
```

a bit lower, just before `while (1)`. Assuming my brain is braining correctly.

dickenhobelix:
Sure, maybe doing the pattern for all LEDs available (i.e. Rx and tx for all channels at the same time) would be a good variation of that.

While not taking additional time, it briefly flashes all of them, i.e. for inspection at the manufacturing line (or at for whatever reason besides looking nice this was originally added)

I'll send a v2 tomorrow

fenugrec:
> Sure, maybe doing the pattern for all LEDs available (i.e. Rx and tx for all channels at the same time) would be a good variation of that.

That works too.

> The blink code is already executed directly before the start of the mainloop

yes indeed, but I had in mind moving that blink code outside the for loop entirely. But disregard this if you go with a "blink all the LEDs of all channels". I'm ok with either

dickenhobelix:
> > Sure, maybe doing the pattern for all LEDs available (i.e. Rx and tx for all channels at the same time) would be a good variation of that.
> 
> That works too.

I just realized that just the hardwired LEDRX and LEDTX is blinked in the original code, as for now I think keeping it that way is probably best. 

> > The blink code is already executed directly before the start of the mainloop
> 
> yes indeed, but I had in mind moving that blink code outside the for loop entirely. But disregard this if you go with a "blink all the LEDs of all channels". I'm ok with either

I had a short glance at the led_sequence API in led.h, which is probably what you have in mind? At the moment the API is not a good fit for flashing the leds alternately, sinde it exercises all LEDs for a given port to the the same value. This could be hacked around by setting RX highactive and TX loactive and resetting anything to default after the init blink sequence, but that does not feel right... Also we would have to introduce a wait loop anyway for stepping forward the blink sequence, which I think is overcomplicating things here, and since @marckleinebudde mentioned in #190 that he wants to rework LED support anyway I think that just exercising the Pins via the HAL API as of the original code is probably ok for now.

I nevertheless cleaned up the patch and just moved the wakeup sequence directly before starting into the mainloop and forcepushed into this branch

src/main.c
```diff
@@ -86,14 +86,6 @@ int main(void)
...
 				 LEDTX_GPIO_Port, LEDTX_Pin, LEDTX_Active_High);
 
-		/* nice wake-up pattern */
-		for (uint8_t j = 0; j < 10; j++) {
-			HAL_GPIO_TogglePin(LEDRX_GPIO_Port, LEDRX_Pin);
-			HAL_Delay(50);
-			HAL_GPIO_TogglePin(LEDTX_GPIO_Port, LEDTX_Pin);
-		}
-
-		led_set_mode(&channel->leds, LED_MODE_OFF);
```
fenugrec, line 96: why not leave the `led_set_mode()`call in here ? the even number of toggles should restore those LEDs to their original state, no ?

actually, led_init() should be setting that state to OFF anyway so it's probably not even necessary at all

Issue #207, opened 2025-12-17
Internal list structure corruption in heavy USB bus load


csanchezdll:
### Summary

I have detected an internal double-linked list corruption problem when candleLight_fw is used on (very) heavily-loaded USB buses. The symptom is a device unable to send further CAN frames; CAN frame reception still works. The only solution is a full FW reset (using the debugger if attached, or power-cycling the device otherwise).

This was observed in a custom STM32G0XX device, but from the analysis (see below) I suspect it will happen in most (all?) devices.

### How to replicate

This only happens in very heavily loaded USB busses where a lot of frames accumulate on the device waiting for sending to the host. As this is not easy to replicate for study, forcing this situation requires some temporary code changes. Also, 2 USB2CAN devices are required (call it DUT and AUX).
**Step 1:** flash DUT with a slightly modified version of candleLight_fw, so sending frames to the host can be disabled using the debugger to simulate a busy USB bus. Add an optional early return to `USBD_GS_CAN_SendToHost`:
```
volatile int skip_send = 0;
void USBD_GS_CAN_SendToHost(USBD_HandleTypeDef *pdev)
{
	if (skip_send)
		return;
[...]
```

**Step 2:** connect the 2 USB2CAN devices so AUX can send CAN frames to DUT. Configure both interfaces:
```
# ip link set can_aux type can bitrate 500000
# ip link set can_aux up
# ip link set can_dut type can bitrate 500000
# ip link set can_dut up
```

**Step 3:** Using GDB, check how many buffer are in the free list. If the device is idle, all of them should be there, except one which is ready to accept the next frame coming from the host:
```
(gdb) define count_nexts
  set $count = 0
  set $i = $arg0.next
  while $i != &($arg0)
    set $count = $count + 1
    set $i = $i->next
  end
  p $count
end
(gdb) count_nexts hGS_CAN.list_frame_pool 
$1 = 191
(gdb) p hGS_CAN.from_host_buf 
$2 = (struct gs_host_frame_object *) 0x2000092c <hGS_CAN+592>
```

**Step 4:** Disable sending frames to the host. This _simulates_ a busy USB bus condition:
```
(gdb) set skip_send = 1
(gdb) continue
```

**Step 5:** Send a number of CAN frames using AUX device equal to the number of buffers in the free list. This causes the available buffer list to become exhausted. Then, try to send two frames using the DUT.
```
$ for i in `seq 1 191`; do cansend can_aux 123#1122334455667788; done
[...]

PR #209, opened 2025-12-26
gs_usb: reserve pre-allocated RX buffers.

csanchezdll:
This PR pre-allocates a set of RX buffers, instead of using a single one in `list_from_host`. This is required in targets using double buffering, because USB HW does not automatically disable the reception of OUT packets, so even when we realize we have no buffers left, the HW might already be receiving a new OUT frame.

With this change, a number (by default 2 on targets with double buffer, 1 on others) of RX buffers are pre-allocated. When pre-allocation is not possible, RX is disabled temporarily. Even if an OUT frame is already incoming, the pre-allocated buffers ensure the callback will have a buffer to put that frame into, to avoid dropping.

Tested on a custom STM32G0 device. Not tested on non-G0 hardware, because I do not have it.

Fixes https://github.com/candle-usb/candleLight_fw/issues/207

### DIFF
--- a/src/usbd_gs_can.c
+++ b/src/usbd_gs_can.c
@@ -613,10 +613,37 @@ static uint8_t USBD_GS_CAN_DataIn(USBD_HandleTypeDef *pdev, uint8_t epnum) {
 	return USBD_OK;
 }
 
+// Prefill the buffers from RX from host.
+// Return true if managed to fill the pre-allocate buffer list
+// and RX can be enabled, false otherwise.
+// Must be called with IRQ disabled.
+static bool USBD_GS_Prefill_RX_Buffers(USBD_GS_CAN_HandleTypeDef *hcan)
+{
+	unsigned i;
+
+	for (i = 0; i < ARRAY_SIZE(hcan->from_host_buf); ++i) {
+		if (hcan->from_host_buf[i])
+			continue;
+
+		hcan->from_host_buf[i] = list_first_entry_or_null(&hcan->list_frame_pool,
+														  struct gs_host_frame_object,
+														  list);
+		if (!hcan->from_host_buf[i])
+			return false;
+
+		list_del(&hcan->from_host_buf[i]->list);
+	}
+
+	return true;
+}
+
 // Note that the return value is completely ignored by the stack.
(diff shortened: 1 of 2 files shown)


src/usbd_gs_can.c
```diff
@@ -613,53 +613,104 @@ static uint8_t USBD_GS_CAN_DataIn(USBD_HandleTypeDef *pdev, uint8_t epnum) {
...
+// Prefill the buffers from RX from host.
+// Return true if managed to fill the pre-allocate buffer list
+// and RX can be enabled, false otherwise.
+// Must be called with IRQ disabled.
+static bool USBD_GS_Prefill_RX_Buffers(USBD_GS_CAN_HandleTypeDef *hcan)
+{
+	unsigned i;
+
+	for (i = 0;
+		 i < sizeof (hcan->from_host_buf) / sizeof(hcan->from_host_buf[0]);
```
fenugrec: we have a macro ARRAY_SIZE for this

src/usbd_gs_can.c
```diff
@@ -613,53 +613,104 @@ static uint8_t USBD_GS_CAN_DataIn(USBD_HandleTypeDef *pdev, uint8_t epnum) {
...
-												   struct gs_host_frame_object,
-												   list);
-	if (hcan->from_host_buf) {
-		list_del(&hcan->from_host_buf->list);
+	list_add_tail(&hcan->from_host_buf[0]->list, &channel->list_from_host);
+
+	unsigned last = 0;
+#if USBD_GS_CAN_RX_BUFFER_COUNT > 1
+	for (last = 0;
+		 last < (sizeof (hcan->from_host_buf) / sizeof(hcan->from_host_buf[0])) - 1;
```
fenugrec: other occurence for ARRAY_SIZE

csanchezdll:
When testing this extensively, I found another bug (https://github.com/candle-usb/candleLight_fw/issues/211). It needs to be fixed before merging this, because otherwise we would be merging code with known frame loses.
I am preparing another PR.

marckleinebudde:
Thanks for the info. I rebased this onto current master, as it had conflicts due to your last fix.

Issue #211, opened 2026-01-08
Frames from host potentially received in wrong msgbuf


csanchezdll:
### Summary

`USBD_GS_CAN_PrepareReceive` is called with interrupts enabled. This is apparently done on purpose (on many places in the code, interrupts are re-enabled just before calling `USBD_GS_CAN_PrepareReceive`; my guess it this was deemed safe because there are no list management calls therein).

The problem is inside `USBD_GS_CAN_PrepareReceive` the app-side msgbuf where next frame is to be received (`from_host_buf`) is prepared using the HAL (through a call to `USBD_LL_PrepareReceive`). If an RX interrupt triggers in the middle, `USBD_GS_CAN_DataOut` can be called, which changes `from_host_buf`; this might cause the HAL to copy next received frame into the wrong msgbuf.

### How to replicate

I found this while extensively testing https://github.com/candle-usb/candleLight_fw/pull/209 (which is itself a fix for a different bug: https://github.com/candle-usb/candleLight_fw/issues/207)

I was unable to find a totally deterministic way to reproduce this. The way I caught this happening was instrumenting the DataOut callback to break when a frame received form the host indicated a channel != 0. As in my tests I was using only channel 0, this indicated the received frame was incorrect (or, more exactly, the received frame had been written somewhere else, and the DataOut callback was trying to process an invalid frame):

```
volatile bool dummy = false;
// Note that the return value is completely ignored by the stack.
static uint8_t USBD_GS_CAN_DataOut(USBD_HandleTypeDef *pdev, uint8_t epnum) {
	USBD_GS_CAN_HandleTypeDef *hcan = (USBD_GS_CAN_HandleTypeDef*)pdev->pClassData;
	if (hcan->from_host_buf[0]->frame.channel != 0)
	{
		dummy = true;
	}
[...]
```

### Analisys

Simplified `USBD_GS_CAN_PrepareReceive` flow:
1. Get location of next frame to be received `from_host_buf`.
2. Call HAL to configure buffer to get next RX to the location obtained in 1.

Simplified `USBD_GS_CAN_DataOut` flow:
1. Enqueue frame just received `from_host_buf` into `list_from_host`.
2. Get new msgbuf into `from_host_buf`.
3. Call  `USBD_GS_CAN_PrepareReceive` (with the 2 steps described above).

[...]

PR #215, opened 2026-01-20
test: new prototypical semi-automatic test.

csanchezdll:
This is the test script I used to detect (and check the fix) of https://github.com/candle-usb/candleLight_fw/issues/207 and https://github.com/candle-usb/candleLight_fw/issues/211.

The README.md should explain everything. Having to build "instrumented" versions of the FW is somewhat clumsy, but is the best way I found of forcing the runtime conditions I wanted to test. Using a breakpoint with and returning inside GDB would be another option, but it is more intrusive in terms of runtime timing.

If you run this before https://github.com/candle-usb/candleLight_fw/pull/209, it will fail right away, as the device will HardFault (trying to receive into a NULL buffer). If you run if on a branch without https://github.com/candle-usb/candleLight_fw/pull/212 (but with https://github.com/candle-usb/candleLight_fw/pull/209, otherwise the more frequent one hides the least frequent one) it will pass, usually, but running in a loop makes it eventually fail (in my system, it takes 100-200 iterations).

No failures running this on main up to 2500 iterations.


test/README.md
```diff
@@ -0,0 +1,55 @@
+# candleLight_gsusb tests #
+
+The script on this directory performs a limited test for
+device functionality. It is meant to be run with a device
+under test (DUT) connected thought USB and JTAG, and
```
fenugrec: JTAG/SWD probably ?

test/README.md
```diff
@@ -0,0 +1,55 @@
...
+another CAN interface (AUX) used to send stimulus to the
+DUT through the CAN bus. Therefore, DUT and AUX must both
+* be visible in as system interfaces
+* be connected ot the same CAN bus
+
+The test is sligtly intrusive, because it temporarily
+suspends the FW capability of sending frames to the host
+thorugh USB. This is done by building a special "test"
```
fenugrec: *through

test/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Running the test ##
+
+1. Prepare your HW setup.
+    * DUT and AUX connected to the same CAN bus
+	* DUT connected to the host by JTAG
+	* GDB server (openocd, most likely) running
+
+2. Start by building the special test version of the
+   code
```
fenugrec: " (append `_test_fw` suffix to your usual target name) "

test/test.sh
```diff
@@ -0,0 +1,151 @@
...
+tmpdir=/tmp/`basename $0`.$PPID
+rm -rf $tmpdir
+mkdir -p $tmpdir
+
+gdb_in=$tmpdir/gdb_in
+gdb_out=$tmpdir/gdb_out
+candump_dut=$tmpdir/candump_dut
+candump_aux=$tmpdir/candump_aux
+
+mkfifo $gdb_in
```
fenugrec: I don't see a corresponding close/delete for this  fifo ?

fenugrec (commented):
some clever bash work in there.
Have not tested, just skimmed; with a few comments.

src/usbd_gs_can.c
```diff
@@ -871,8 +836,16 @@ static uint8_t USBD_GS_CAN_SendFrame(USBD_HandleTypeDef *pdev, struct gs_host_fr
...
+#ifdef USBD_TEST
+volatile int skip_send = 0;
+#endif
 void USBD_GS_CAN_SendToHost(USBD_HandleTypeDef *pdev)
 {
+#ifdef USBD_TEST
+	if (skip_send)
+		return;
+#endif
+
```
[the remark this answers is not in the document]
csanchezdll: I am pretty sure static does not affect visibility in GDB, but I am checking. About init, you are totally right. I'll use IS_ENABLED()... or should we just remove USBD_TEST altogether and avoid splitting the FW binaries?

test/test.sh
csanchezdll: Yup, I know, that is what I am doing. This script started as a quick & dirty thing to catch the bug, that is why the "failing" exit was not so taken care of.

test/test.sh
```diff
@@ -0,0 +1,180 @@
...
+    if [ -n "$gdb_pid" ]
+    then
+	kill -INT $gdb_pid
+	echo "quit" > $gdb_in
+    fi
+    if $keep_tmpdir
+    then
+	printf "Keeping temporary directory in %s\n" $tmpdir
+    else
+	rm -rf $tmpdit
```
csanchezdll: tmpdir

PR #247, opened 2026-03-19
cleanup EP0 buffer size calculation

marckleinebudde:
using these `maxX` constructs is too error prone.

### DIFF
--- a/include/dfu.h
+++ b/include/dfu.h
@@ -26,4 +26,13 @@ THE SOFTWARE.
 
 #pragma once
 
+#include <stdint.h>
+
+struct dfu_status {
+	uint8_t status;
+	uint8_t poll_timeout[3];
+	uint8_t state;
+	uint8_t stringidx;
+};
+
 void dfu_run_bootloader(void);
--- a/include/usbd_gs_can.h
+++ b/include/usbd_gs_can.h
@@ -32,6 +32,7 @@ THE SOFTWARE.
 #include "can.h"
 #include "compiler.h"
 #include "config.h"
+#include "dfu.h"
 #include "gs_usb.h"
 #include "led.h"
 #include "list.h"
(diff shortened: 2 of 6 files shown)


include/dfu.h
```diff
@@ -26,4 +26,13 @@ THE SOFTWARE.
 
 #pragma once
 
+#include <stdint.h>
+
+struct dfu_status {
```
fenugrec, line 31: no `__packed__` directive on this ?

include/usbd_gs_can.h
```diff
@@ -83,7 +84,23 @@ struct gs_host_frame_object {
 };
 
 typedef struct {
-	uint8_t __aligned(4) ep0_buf[GS_CAN_EP0_BUF_SIZE];
```
fenugrec, line 86: I recall some hardfaults on F0 that would not always manifest, depending on on the alignment of this struct... does struct_group_* ensure alignment somehow ?

include/dfu.h
marckleinebudde: Doh! Will add.

include/usbd_gs_can.h
marckleinebudde: I'll add some more `__aligned(4)` in there.

include/dfu.h
marckleinebudde: yes, `__packed` but not `__aligned(4)`, as it will increase the size of the struct to `8`.

fenugrec: Sometimes I put a _static_assert(sizeof(....) on a struct if I need to be sure I don't accidentally mess it up later : ) I don,t think we ever did that anywhere in this codebase though ?

marckleinebudde: Harhar, I was using the same to double check `pahole`'s output. Will add a `static_assert(sizeof(struct dfu_status) == 6);`

PR #248, opened 2026-03-20
Cleanup ep0 handling

marckleinebudde:


src/usbd_gs_can.c
```diff
@@ -418,7 +419,7 @@ static uint8_t USBD_GS_CAN_Config_Request(USBD_HandleTypeDef *pdev, USBD_SetupRe
 		case GS_USB_BREQ_DATA_BITTIMING:
 		case GS_USB_BREQ_SET_TERMINATION:
 		case GS_USB_BREQ_SET_FILTER:
-			if (req->wLength > sizeof(hcan->ep0)) {
+			if (req->wLength > sizeof(*ep0)) {
```
fenugrec, line 422: does taking `sizeof` a union always do what we want here ?

include/gs_usb.h
```diff
@@ -225,7 +225,6 @@ enum gs_can_state {
 enum gs_can_termination_state {
 	GS_CAN_TERMINATION_STATE_OFF = 0,
 	GS_CAN_TERMINATION_STATE_ON,
```
fenugrec, line 227: `get_term()` has a nice return type of `enum gs_can_termination_state`; it seems... inelegant to be doing `return -1` from there ?

src/usbd_gs_can.c
marckleinebudde: The functionality didn't change here.

What's the purpose of that `if()`? The host sends us an EP0 config request and if the length of that request is longer than our buffer we receive into, the firmware will throw an error. This effectively allows the host to send more data for a certain request than we need, but I think that's OK.

To answer your question:

> does taking sizeof a union always do what we want here ?

Yes, the union holds enough space for the longest buffer, here a `pahole -a` output of it:

What the macro creates is a little but ugly, but I think it's correct:

```
struct ep0_data {
	union {
		struct dfu_status  dfu_status;           /*     0     6 */
		const struct gs_host_config config;      /*     0     4 */
		const struct gs_device_bittiming bittiming __attribute__((__aligned__(4))); /*     0    20 */
		const struct gs_device_mode mode __attribute__((__aligned__(4))); /*     0     8 */
		const struct gs_identify_mode identify_mode; /*     0     4 */
		const struct gs_device_filter filter __attribute__((__aligned__(4))); /*     0   132 */
		struct gs_device_termination_state term_state; /*     0     4 */
	} __attribute__((__aligned__(4)));               /*     0   132 */

	/* size: 132, cachelines: 3, members: 1 */
	/* forced alignments: 1 */
	/* last cacheline: 4 bytes */
} __attribute__((__aligned__(4)));


union ep0 {
	union {
		struct {
			union {
				struct dfu_status dfu_status; /*     0     6 */
				const struct gs_host_config config; /*     0     4 */
				const struct gs_device_bittiming bittiming __attribute__((__aligned__(4))); /*     0    20 */
				const struct gs_device_mode mode __attribute__((__aligned__(4))); /*     0     8 */
				const struct gs_identify_mode identify_mode; /*     0     4 */
				const struct gs_device_filter filter __attribute__((__aligned__(4))); /*     0   132 */
				struct gs_device_termination_state term_state; /*     0     4 */
			} __attribute__((__aligned__(4)));   /*     0   132 */
		} __attribute__((__aligned__(4))) __attribute__((__aligned__(4)));     /*     0   132 */
		struct ep0_data    data __attribute__((__aligned__(4))); /*     0   132 */
	} __attribute__((__aligned__(4)));             /*     0   132 */
	uint8_t                    buf[132];           /*     0   132 */
};


typedef struct {
	union ep0                  ep0 __attribute__((__aligned__(4))); /*     0   132 */
	/* --- cacheline 2 boundary (128 bytes) was 4 bytes ago --- */
	USBD_SetupReqTypedef       last_setup_request;   /*   132     8 */
	struct list_head           list_frame_pool;      /*   140     8 */
	struct list_head           list_to_host;         /*   148     8 */
	struct gs_host_frame_object * from_host_buf[2];  /*   156     8 */
	struct gs_host_frame_object * to_host_buf;       /*   164     4 */
	can_data_t                 channels[1] __attribute__((__aligned__(4))); /*   168   208 */
	/* --- cacheline 5 boundary (320 bytes) was 56 bytes ago --- */
	uint32_t                   feature;              /*   376     4 */
	uint32_t                   sof_timestamp_us;     /*   380     4 */
	/* --- cacheline 6 boundary (384 bytes) --- */
	struct gs_host_frame_object msgbuf[64] __attribute__((__aligned__(4))); /*   384  2048 */
	/* --- cacheline 38 boundary (2432 bytes) --- */
	_Bool                      dfu_detach_requested; /*  2432     1 */

	/* size: 2436, cachelines: 39, members: 11 */
	/* padding: 3 */
	/* forced alignments: 3 */
	/* last cacheline: 4 bytes */
} USBD_GS_CAN_HandleTypeDef __attribute__((__aligned__(4)));
```

include/gs_usb.h
marckleinebudde: Yes it is. But with not enabled `CONFIG_TERMINATION` it's effectively not used anymore. But you've got a point. I'll add back a special value to the `enum` and return it.

src/usbd_gs_can.c
fenugrec: thanks for clarifying; I agree it should be fine.

marckleinebudde: Thanks for looking into the code.

PR #278, opened 2026-05-22
fix "LOAD segment with RWX permissions" warning

### LINKED ISSUE #106: recent gcc complains about "LOAD segment with RWX permissions"

marckleinebudde:
The linker treats the `.copy.table`, `.preinit_array`, and `.init_array` sections as data, which are writable by default. Explicitly mark them as `READONLY` (on supported linkers) to fix this warning:

```
warning: candleLight_fw has a LOAD segment with RWX permissions
```

Closes: https://github.com/candle-usb/candleLight_fw/issues/106

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,8 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
 	set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "Debug" "Release")
 endif()
 
+####### check for newlib and picolibc
+
 include(CheckSymbolExists)
 
 check_symbol_exists(__PICOLIBC__ "sys/features.h" HAVE_PICOLIBC)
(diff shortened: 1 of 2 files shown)


fenugrec:
huh. Wonder why they were still tagged "RWX", maybe just because they end up in same mem area as other +X sections. Good find!

marckleinebudde:
These sections are in the flash, so the writable attribute is just an academic problem 😸 But I wanted the warning to go away. It only shows up with the external Arm toolchain.

PR #319, opened 2026-09-16
Fix jump from STM bootloader to application

### LINKED ISSUE #195: Firmware restart after USB DFU update fails
When updating the firmware on a CandleLightFD adapter the final restart of the newly flashed firmware does not work:

```
mkorber@raspberrypi:~ $ sudo dfu-util -d 1d50:606f -a 0 -s 0x08000000:leave -D candleLightFD_fw_8d3e7fa_2023-07-20.bin
dfu-util 0.11

Copyright 2005-2009 Weston Schmidt, Harald Welte and OpenMoko Inc.
Copyright 2010-2021 Tormod Volden and Stefan Schmidt
This program is Free Software and has ABSOLUTELY NO WARRANTY
Please report bugs to http://sourceforge.net/p/dfu-util/tickets/

dfu-util: Warning: Invalid DFU suffix signature
dfu-util: A valid DFU suffix will be required in a future dfu-util release
Opening DFU capable USB device...
Device ID 1d50:606f
Run-Time device DFU version 011a
Claiming USB DFU (Run-Time) Interface...
Setting Alternate Interface zero...
Determining device status...
DFU state(0) = appIDLE, status(0) = No error condition is present
Device really in Run-Time Mode, send DFU detach request...
dfu-util: error detaching
Device will detach and reattach...
Opening DFU USB Device...
Claiming USB DFU Interface...
Setting Alternate Interface #0 ...
Determining device status...
DFU state(2) = dfuIDLE, status(0) = No error condition is present
DFU mode device DFU version 011a
Device returned transfer size 1024
DfuSe interface name: "Internal Flash   "
Downloading element to address = 0x08000000, size = 19504
Erase           [=========================] 100%        19504 bytes
Erase    done.
Download        [=========================] 100%        19504 bytes
Download done.
File downloaded successfully
Submitting leave request...
dfu-util: Error during download get_status
```

After that the USB interface did not automatically re-enumerate but it was required to unplug/replug the adapter.

The issue seems to be very similar to https://github.com/candle-usb/candleLight_fw/pull/176#issuecomment-2128253129 So it is probably not related to the CandleLightFD adapter.

Any ideas or input on that issue?

Semiotmo:
Jump from STM bootloader to application is currently broken.

When jumping from STM integrated bootloader to application (e.g. after a firmware update via dfu), SCB->VTOR is still set to point to the bootloaders vector table. 
The STM bootloader does not reset this to the appropriate value on exit. It expects that the application itself sets the correct VTOR, as documented in AN3156, Rev 18, page 20.
> The Jump to application works only if the user application sets the vector table correctly to
point to the application address.

When VTOR is not set to the applications vector table and an interrupt arrives, the application executes the interrupt handler defined in STM bootloader code instead of its own. In my case this crashed the MCU.

This pull request fixes this by making sure that the VTOR is set appropriately and no interrupt can arrive before it is set.

Additionally it ensures that after a successful jump from bootloader to application, the application does not directly jump back to bootloader by explicitly clearing the `dfu_reset_to_bootloader_magic` variable. 
I decided to explicitly do it as well as I don't know if the STM bootloader will reset the memory associated with this variable itself.

This fix enables using the `:leave` parameter of `dfu-util` ( `dfu-util -s 0x08000000:leave ...`).

Closes: https://github.com/candle-usb/candleLight_fw/issues/195


src/dfu.c
```diff
@@ -81,7 +91,15 @@ void __initialize_hardware_early(void)
 		}
 	}
 
+	/* If SCB->VTOR is configured to be not located at FLASH_BASE, this will be overwritten by `SystemInit()`. */
+	SCB->VTOR = FLASH_BASE;
```
marckleinebudde: Please squash so that it compiles on stm32f072.

```suggestion
#if defined(__VTOR_PRESENT)
	/* If SCB->VTOR is configured to be not located at FLASH_BASE, this will be overwritten by `SystemInit()`. */
	SCB->VTOR = FLASH_BASE;
#endif

```

marckleinebudde (changes requested):
Thanks for your contribution!
Works for me on stm32g0b1. Please squash suggested fix.

marckleinebudde:
This will fix https://github.com/candle-usb/candleLight_fw/pull/176#issuecomment-2128253129 and
Closes: https://github.com/candle-usb/candleLight_fw/issues/195

marckleinebudde (approved):

src/dfu.c
marckleinebudde: I've added the changes to your patch.

